PR #15, opened 2023-06-30
Cvr loader

jennjwang:
assertion fails for test_undervote(). I ran the same example in the profile.py file and it returned True, which contradicts the result in the test file.


jennjwang:
ready for review

src/unnamed_rcv_thing/cvr_loader.py
```diff
@@ -0,0 +1,81 @@
+import pandas as pd
+import pathlib
+import os
+from pydantic import BaseModel
+from typing import Callable
+from unnamed_rcv_thing.profile import PreferenceProfile
```
jamesturk, line 6: when importing files from within the package, you can do `from .profile import PreferenceProfile` instead of repeating the package name (makes future refactors easier)

src/unnamed_rcv_thing/cvr_loader.py
```diff
@@ -0,0 +1,81 @@
...
+            ValueError: if the voter id column has missing values
+            DataError: if the voter id column has duplicate values
+
+        Returns:
+            PreferenceProfile: preference schedule that contains all the parsed ballots
+        """
+        if not os.path.isfile(fpath):
+            raise FileNotFoundError(f'File with path {fpath} cannot be found')
+        cvr_path = pathlib.Path(fpath)
+        df = pd.read_csv(cvr_path, on_bad_lines='error', encoding="utf8")
```
jamesturk, line 39: A question I have about this implementation is whether or not the load_func should be wrapped like this.

One option would be to move this logic into rank_column_csv & not assume that all files being loaded are CSVs.

This would mean there's no need for CVRLoader, load functions would be called directly.

src/unnamed_rcv_thing/cvr_loader.py
```diff
@@ -0,0 +1,81 @@
...
+    df = pd.read_csv(cvr_path, on_bad_lines='error', encoding="utf8")
+
+    ranks = list(df.columns)
+    if id_col is not None:
+        ranks.remove(df.columns[id_col])
+    grouped = df.groupby(ranks, dropna=False)
+    ballots = []
+
+    for group, group_df in grouped:
+        group = [None if pd.isnull(c) else c for c in group]
```
jamesturk, line 71: minor thing, but could save the unnecessary casts to a list and second iteration by combining these into something similar to:

ranking = [{None} if pd.isnull(c) else {c} for c in group]

src/unnamed_rcv_thing/profile.py
```diff
@@ -39,17 +40,11 @@ def get_candidates(self):
...
-    #         ballots (list of Ballot): a list of ballots in the election
-    #         candidates (list of Candidates): a list of candidates in the election
-    #     """
-    #     self.id = uuid.uuid4()
-    #     self.ballots = ballots
-    #     self.candidates = candidates
-    #     self.ballot_weights = [b.score for b in ballots]
+        
+
+if __name__ == '__main__':
```
jamesturk, line 45: I'd move this to a test instead of here so it gets run regularly & doesn't wind up diverging from the code over time.

jamesturk (commented):
Looks good, left some comments for potential areas to reconsider.

jamesturk:
👍

PR #24, opened 2023-07-18
Cleaning Rule for Deduplication

jennjwang:


jennjwang:
should we allow rankings in the same profile of ballots to have different lengths?

jamesturk:
Hi Jenn,

This PR has a lot of unrelated changes & now some conflicts with main, any chance you all can talk through those in person to try to get this back into a reviewable/mergable state? 

In the meantime, if there's a piece you want me to review let me know what portion of the files represents the actual PR since all the extra changes made it unclear. Thanks

tests/test_models.py
```diff
@@ -1,28 +1,28 @@
-import pytest
-from votekit.models import Outcome
+# import pytest
```
jamesturk, line 1: if these tests aren't meant to pass they should either be removed or marked with pytest.mark.skip
https://docs.pytest.org/en/7.1.x/how-to/skipping.html

tests/data/.DS_Store
jamesturk, line 1: add .DS_Store to .gitignore to avoid littering the repo with these (macOS specific) files

tests/data/csv/alaska.csv
jamesturk, line 1: careful with files this large, this is how the other repo got to be so big in the first place.  a couple are fine, but in general try to use the smallest file possible to demonstrate something with a test

src/votekit/cvr_loaders.py
jamesturk, line 1: since this library isn't used yet this reordering of parameters is fine, but generally you only want to add parameters at the end of a list, (this change would have broken code that expected the second element to be id_col). 

You can also use this syntax:

`def rank_column_csv(fpath, *, weight_col, delimiter, id_col)`

To make the arguments after the `*` keyword only, so people providing them will have to specify which is which and not rely on order.  This leads to more readable code at the expensive of a tiny bit more typing/autocomplete :)

src/votekit/cleaning.py
```diff
@@ -20,3 +25,90 @@ def remove_empty_ballots(
         pp_clean = PreferenceProfile(ballots=ballots_nonempty)
 
     return pp_clean
+
+
+def clean(
```
jamesturk, line 30: If this is meant to only be used internally, consider renaming to _clean so a person browsing the code would take the hint that it isn't meant to be used by library users.

tests/test_cleaning.py
```diff
@@ -15,3 +15,106 @@ def test_remove_empty_ballots():
...
+    cleaned = PreferenceProfile(
+        ballots=[
+            Ballot(
+                ranking=[{"C", "A"}, {"B"}], weight=Fraction(2), voters={"tom", "andy"}
+            )
+        ]
+    )
+    assert cleaned == profile_res
+
+
```
jamesturk, line 105: tests look good!

jamesturk (commented):
code looks good overall, left a few comments on possible changes

jamesturk:
As for the question about different length ballots, are you aware of any reason why it'd happen?  & Right now how does the code handle it?

tests/test_models.py
jennjwang: uncommented, tests are passing but thanks for the note!

tests/data/csv/alaska.csv
jennjwang: ok I'll remove it

jennjwang:
> As for the question about different length ballots, are you aware of any reason why it'd happen? & Right now how does the code handle it?

It can happen during deduplication (ex. [a, a, b, c] can become [a, b, c] or [a, None, b, c]). I think we'll probably have two deduplicate functions to handle both cases.

jamesturk:
Gotcha, that makes sense.  Depending on implementation might make sense as configuration to deduplication function.

PR #26, opened 2023-07-19
stv run election w/tests

jgibson517:
Hi James,

This PR has the fully implemented STV class with an initial batch of tests. One question I had regarding tests for election, since were assuming preference profiles will be cleaned, do you think its worth writing tests that handles profiles with missing ballots, or 'None' candidates?

Also, I cleaned up and added a test for the to_dict() method in the preference profile, which can be reviewed now too!

Thanks!


src/votekit/profile.py
```diff
@@ -55,14 +55,14 @@ def num_ballots(self) -> Fraction:
 
     def to_dict(self) -> dict:
         """
-        Converts balots to dictionary with keys, ranking and
+        Converts balots to dictionary with keys, ranking, tuple and
```
jamesturk, line 58: small typo "bal[l]ots" 

also this docstring could be more clear, I had to read it a few times while looking at the code to understand what it was saying.

Maybe something like "Converts ballots to a dict mapping rankings to weights"

tests/test_profile.py
```diff
@@ -14,3 +24,19 @@ def test_unique_cands():
...
     assert len(cands) == len(unique_cands)
+
+
+def test_updates_in_place():
+    before = test_profile.get_ballots()
+    remove = "a"
+    after = remove_cand(remove, before)
+    assert remove in test_profile.get_candidates()
+    for ballot in after:
+        if remove in ballot.ranking:
```
jamesturk, line 35: `if remove in ballot.ranking: assert remove in ballot.ranking` is never going to fail, worth revisiting to decide what you actually want to test here

src/votekit/election_types.py
```diff
@@ -55,19 +52,18 @@ def run_step(self, profile: PreferenceProfile) -> tuple[PreferenceProfile, Outco
...
                 ballots = self.transfer(candidate, ballots, fp_votes, self.threshold)
 
-        if not self.is_complete():
+        if self.next_round():
             lp_votes = min(fp_votes.values())
             lp_candidates = [
                 candidate for candidate, votes in fp_votes.items() if votes == lp_votes
             ]
             # is this how to break ties, can be different based on locality
+            print(f"Eliminated: {lp_candidates}")
```
jamesturk, line 63: stray print statement?

jamesturk (changes requested):
core functionality looks good, a couple of small things to address

src/votekit/cleaning.py
jgibson517, line 1: Sorry this file does not need review, one of my commits slipped into the PR

PR #27, opened 2023-07-19
updated outcome_model, test_outcome_model, ballot_graph

MalavikaMukundan:


src/votekit/ballot_graph.py
```diff
@@ -0,0 +1,226 @@
...
+import networkx as nx
+from collections import Counter
+from profile import PreferenceProfile
+from distinctipy import get_colors
+from ballot import Ballot
+from typing import Optional, Callable
+from pydantic import BaseModel, validator
+
+
+def build_graph(n):
```
jamesturk, line 11: worth running `ruff` and `black` on this to bring it in line w/ style guidelines, not going to flag all the issues here but will mention a few worth paying special attention to

src/votekit/ballot_graph.py
```diff
@@ -0,0 +1,226 @@
...
+            # reduce everything mod new_label
+        for i in range(len(tmp)):
+            if tmp[i]> num_cands:
+                tmp[i] = tmp[i]- num_cands
+        node_map[k] = tuple([new_label]+tmp)
+
+    return nx.relabel_nodes(gr, node_map)
+
+
+Graphs = {}
```
jamesturk, line 83: this global variable seems to be used for caching, it should either be labeled with a comment justifying its usage or replaced with something intended for caching (see functools.lru_cache)

src/votekit/ballot_graph.py
```diff
@@ -0,0 +1,226 @@
...
+
+
+class BallotGraph():
+    profile: PreferenceProfile
```
jamesturk, line 87: defined this way, these are class-level variables, shared by instances- which I don't think is the intention here. consider just declaring these within __init__, or making this class inherit from pydantic.BaseModel if that was the intent

src/votekit/ballot_graph.py
```diff
@@ -0,0 +1,226 @@
...
+                self.ballot_dict.pop(ballot)
+            elif len(ballot)==self.num_cands-1:
+                for i in self.profile.get_candidates():
+                    if i not in ballot:
+                        self.ballot_dict[ballot+tuple([i])]+= di[ballot]    
+                        self.ballot_dict.pop(ballot)
+                        break
+                
+        
+    def visualize(self, neighborhoods = {}):
```
jamesturk, line 135: passing a dictionary as a default argument means it will be reused on each call, which I don't believe is the intention here

see https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

src/votekit/ballot_graph.py
```diff
@@ -0,0 +1,226 @@
...
+                if self.ballot_dict[bal]!=0:
+                    i = self.profile.get_candidates().index(bal[0])
+                    node_cols.append(cols[i])
+                else:
+                    node_cols.append(WHITE)
+            ##want to include number of votes as part of labels,  color ballots with 0 votes grey
+        
+            nx.draw_networkx(Gc,with_labels = True, node_color = node_cols)
+            
+        else:
```
jamesturk, line 155: lots of repeated code between these two conditionals, that code should be moved outside the conditionals so it isn't repeated, with only the parts that differ within the if/else blocks

src/votekit/ballot_graph.py
```diff
@@ -0,0 +1,226 @@
...
+            centers = list(neighborhoods.keys())
+            
+            for bal in Gc.nodes:
+                found = False
+                for i in range(len(centers)):    
+                    weight = (neighborhoods[centers[i]])[1]
+                    if bal in (neighborhoods[centers[i]])[0].nodes:
+                        node_cols.append(tuple([(1-self.ballot_dict[bal]/weight)*x for x in cols[i]]))
+                        found = True
+                        break ##breaks the inner for loop
```
jamesturk, line 169: comments should focus on the "why" not the "what", a person reading this would presumably know that break exits a for loop, explaining why that's happening is more useful/important than saying what break does here

src/votekit/ballot_graph.py
```diff
@@ -0,0 +1,226 @@
...
+        
+    @staticmethod
+    def show_all_ballot_types(n):
+        if n not in Graphs.keys():
+            build_graph(n)
+        Gc = Graphs[n]
+        nx.draw(Gc, with_labels = True)
+        plt.show()
+  
+    def compare(self, new_pref: profile, dist_type: Callable):
```
jamesturk, line 187: consider removing or raising NotImplemented for now so calls will fail

src/votekit/outcome_model.py
```diff
@@ -0,0 +1,140 @@
...
+from profile import PreferenceProfile
+from typing import Optional
+
+#1. "--> int" arrow thing for each function
+#2. is add_winners_and_losers necessary? --A: probabably not
+#3. do we want anything more?]
+#4. RCV cruncher: inspiration for funcitons--> statistics functions/things
+# Example of immutable data model for results
+
+class Outcome():
```
jamesturk, line 11: this should inherit from pydantic.BaseModel if that's the intention, later on pydantic Config is used, but inheritance is missing

src/votekit/outcome_model.py
```diff
@@ -0,0 +1,140 @@
...
+        self.eliminated = eliminated
+        self.remaining = remaining
+        self.profile = profile
+        self.winner_votes = winner_votes
+        self.previous = previous
+
+    class Config:
+        allow_mutation = False
+
+    def get_curr_round(self) -> int:
```
jamesturk, line 34: there is no need to write accessors like this in Python, just have people access the variables directly if you're just returning them

src/votekit/outcome_model.py
```diff
@@ -0,0 +1,140 @@
...
+        return Outcome(
+            remaining=set(self.remaining).difference(winners | losers),
+            elected=list(set(self.elected) | winners)
+            eliminated=list(set(self.eliminated) | losers)
+        )
+    """
+    
+
+    
+    # THIS ASSUMES BALLOTS DON'T CHANGE BY ROUND!
```
jamesturk, line 99: assumptions like this should be documented in the docstring and/or enforced via error checking

tests/test_outcome_model.py
```diff
@@ -0,0 +1,98 @@
+import sys
+sys.path.insert(0, r"C:\Users\malav\OneDrive\Desktop\mggg\VoteKit-Mala\src\votekit")
```
jamesturk, line 2: this shouldn't be here as it will only work on one computer

jamesturk (changes requested):
started reviewing this one before realizing it wasn't in a complete state, but still left some notes that might be helpful.

as a tip, if something isn't ready for review, hold off on PR or mark the PR as pending

src/votekit/outcome_model.py
jgibson517, line 99: Ballots change over round, so the first conditional is redundant

src/votekit/outcome_model.py
```diff
@@ -0,0 +1,140 @@
...
+        Returns dict of (key) string candidates who changed ranking from previous round and (value) a tuple of (prevRank, newRank)
+        """
+        if self.previous == None:
+            raise ValueError("This is the first round, cannot compare previous ranking")
+        
+        else:
+            prev_ranking = self.previous.get_rankings()
+            curr_ranking = self.get_rankings()
+            if curr_ranking == prev_ranking:
+                return {}
```
jgibson517, line 124: Let's try and avoid returning empty dictionaries, maybe this is a print(statement) that states the current and previous rankings are the same

src/votekit/ballot_graph.py
jgibson517, line 11: Let's add some doc-strings to this, also is there a reason the live outside the class?

src/votekit/ballot_graph.py
```diff
@@ -0,0 +1,226 @@
...
+                for i in range(len(centers)):    
+                    weight = (neighborhoods[centers[i]])[1]
+                    if bal in (neighborhoods[centers[i]])[0].nodes:
+                        node_cols.append(tuple([(1-self.ballot_dict[bal]/weight)*x for x in cols[i]]))
+                        found = True
+                        break ##breaks the inner for loop
+                if not found:
+                    node_cols.append(WHITE)
+            nx.draw_networkx(Gc, with_labels = True, node_color = node_cols)
+        return
```
jgibson517, line 173: either return a plot object or delete the return statement

src/votekit/ballot_graph.py
```diff
@@ -0,0 +1,226 @@
...
+                        self.ballot_dict[ballot+tuple([i])]+= di[ballot]    
+                        self.ballot_dict.pop(ballot)
+                        break
+                
+        
+    def visualize(self, neighborhoods = {}):
+        Gc = Graphs[self.num_cands]    
+        if neighborhoods == {}:
+            self.clean()
+            WHITE = (1,1,1)
```
jgibson517, line 139: Will these WHITE/BLACK variables be the same for every instance a visualized graph or is that something the user should be able to input?

src/votekit/ballot_graph.py
jgibson517, line 83: What is this dictionary and why is outside of a class/function? is this just a place to store any created graphs? Will a user need to access it/modify it/create it?

jgibson517 (changes requested):
This looks good! left some notes throughout. If you have poetry installed, you could run `poetry run pre-commit install` which will run some formatting checks before you commit/push!

jamesturk:
marking as draft since this is still WIP, unmark as draft / comment when ready

src/votekit/ballot_graph.py
MalavikaMukundan: Deleted the variable Graphs

```suggestion
@cache
def build_graph(n):
    print("Called on" + str(n))
    Gc = nx.Graph()
```

src/votekit/ballot_graph.py
MalavikaMukundan:
```suggestion
class BallotGraph():

    def __init__(self,profile):

```

MalavikaMukundan: Changed to just declaring them within init

src/votekit/ballot_graph.py
MalavikaMukundan: Changed {} to None and modified the if condition that follows

src/votekit/ballot_graph.py
MalavikaMukundan:
```suggestion
   def visualize(self, neighborhoods:dict = None):##visualize the whole election or select neighborhoods in the election. TODO: change this so that neighborhoods can have any neighborhood, not just heavy balls 
        Gc = build_graph(self.num_cands)  
        WHITE = (1,1,1)
        BLACK=(0,0,0)
        node_cols = []
        
        if neighborhoods==None:
            k = self.num_cands
        else:
            k = len(neighborhoods)
            
        cols = get_colors(k, [WHITE, BLACK])
        
        self.clean()
        for bal in Gc.nodes:
            i=-1
            color=WHITE
            weight = self.num_voters
            
            if neighborhoods==None:
                if self.ballot_dict[bal]!=0:
                    i=(self.profile.get_candidates()).index(bal[0])
                    
            else:
                for c in neighborhoods.keys():
                    if bal in (neighborhoods[c])[0].nodes:
                        weight = (neighborhoods[c])[1]
                        i = (list(neighborhoods.keys())).index(c)
                        break
            if i!=-1:
                color = tuple([(1-self.ballot_dict[bal]/weight)*x for x in cols[i]])
            
            node_cols.append(color)
        
        nx.draw_networkx(Gc, with_labels = True, node_color = node_cols)
        plt.show()
        return
```

src/votekit/outcome_model.py
MalavikaMukundan: Deleted

src/votekit/outcome_model.py
```diff
@@ -0,0 +1,140 @@
...
+            return l
+    def get_rankings(self) -> list[str]:
+        return self.get_all_winners() + self.remaining +  self.get_all_eliminated()
+    
+    def get_profile(self) -> PreferenceProfile:
+        if self.profile != None:
+            return self.profile
+        elif self.previous==None:
+            raise ValueError(f"No profile found")
+        else:
```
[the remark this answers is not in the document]
MalavikaMukundan: The profile (overall ballot counts) is just something we'd probably feed into Outcome when we define the first round outcome; we wouldn't add it to the later rounds. So when you're accessing the outcome at a later round you wanna go back rounds until you find it (if it's been given to Outcome at all)

src/votekit/outcome_model.py
```diff
@@ -0,0 +1,140 @@
...
+            return self.previous.get_profile()
+    
+    def get_round_outcome(self,roundNum: int) -> list[str]:
```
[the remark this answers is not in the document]
MalavikaMukundan: def get_round_outcome(self,roundNum: int) -> dict:
        if self.curr_round == roundNum:
            return {'elected':self.elected, 'eliminated':self.eliminated}
        else:
            self.previous.get_round_outcome(roundNum)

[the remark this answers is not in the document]
MalavikaMukundan:
```suggestion
def get_round_outcome(self,roundNum: int) -> dict:
        if self.curr_round == roundNum:
            return {'elected':self.elected, 'eliminated':self.eliminated}
        else:
            self.previous.get_round_outcome(roundNum)
    
```

src/votekit/ballot_graph.py
MalavikaMukundan: no they're gonna be the same for every instance. They just specify the colors that you don't want to use when visualizing. White doesn't make sense, and using black would mess with the visibility of the node labels

src/votekit/ballot_graph.py
MalavikaMukundan: It was supposed to be used for caching, but I've deleted it now and cached build_graph instead

src/votekit/outcome_model.py
```diff
@@ -0,0 +1,140 @@
...
+    
+    def get_all_winners(self) -> list[str]: # returns in order of first elected - last elected
+        if self.previous != None:
+            return self.previous.get_all_winners() + self.elected
+        else:
+            return self.elected
+    
+    def get_all_eliminated(self) -> list[str]: #returns in order of last eliminated - first eliminated
+        if self.previous != None:
+            l = self.eliminated.copy()
```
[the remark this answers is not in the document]
MalavikaMukundan: Done

src/votekit/outcome_model.py
MalavikaMukundan: gotcha

src/votekit/ballot_graph.py
MalavikaMukundan: Added doc-string. I want to keep it outside so we can use it even if we're not initializing an instance using BallotGraph. Open to alternate suggestions, but I still wanna cache it

src/votekit/ballot_graph.py
MalavikaMukundan:
```suggestion
        nx.draw_networkx(Gc, with_labels = True, node_color = node_cols)
        plt.show()
        
```

MalavikaMukundan:
Will open another one later

PR #28, opened 2023-07-19
undervote function check 

edelanuez:
### this is the undervote function that truncates ballots with empty rankings


jamesturk:
marking as draft for now since this appears to be WIP, mark as ready for review/comment when ready

PR #32, opened 2023-07-20
All work so far since Monday 7/17

roryhe:
This includes random transfer, blt parser, testing (including running 17 Scottish elections with STV class), and documentation.


tests/test_load_blt.py
```diff
@@ -0,0 +1,82 @@
...
+
+    election = STV(mn_profile, fractional_transfer, seats=1)
+
+    while not election.is_complete():
+        mn_profile, out = election.run_step(mn_profile)
+
+    print(out)
+
+
+def test_blt():
```
jgibson517, line 25: what are these two tests testing?

tests/test_quota.py
```diff
@@ -0,0 +1,53 @@
+from votekit.cvr_loaders import blt
+from votekit.election_types import STV, fractional_transfer
+from pathlib import Path
+import pytest
+
+BASE_DIR = Path(__file__).resolve().parent
+DATA_DIR = BASE_DIR / "data"
+
+
+def test_droop_default_parameter():
```
jgibson517, line 10: the threshold tests should check that values computed are correct, not the election itself, here's some pseudo code to show what I mean 
`
def test_hare(): 
    election = STV(pp, fractional_transfer, seats=seats, quota="Droop")
    actual = 34.89
    assert threshold == election.threshold

`

src/votekit/cvr_loaders.py
```diff
@@ -68,3 +69,89 @@ def rank_column_csv(
...
+    cands_included = False
+
+    if not os.path.isfile(fpath):
+        raise FileNotFoundError(f"File with path {fpath} cannot be found")
+    if os.path.getsize(fpath) == 0:
+        raise EmptyDataError("Dataset cannot be empty")
+
+    with open(fpath, "r") as file:
+        for i, line in enumerate(file):
+            s = line.rstrip("\n").rstrip()
```
jgibson517, line 105: why are we striping the line twice here?

src/votekit/cvr_loaders.py
```diff
@@ -68,3 +69,89 @@ def rank_column_csv(
...
+                ballots = [
+                    Ballot(
+                        ranking=[{name_map[cand]} for cand in ballot[0]],
+                        weight=Fraction(ballot[1]),
+                    )
+                    for ballot in ballots
+                ]
+
+                return PreferenceProfile(ballots=ballots, candidates=names), seats
+    raise Exception(f"Undetermined error parsing file '{fpath}'.")
```
jgibson517, line 157: this is probably redundant given the thoroughness of your other checks + python's built in errors will probably be raised during parsing before this gets triggered

jgibson517 (changes requested):
left some comments throughout, let me know if you have any questions. for testing files, lets include a shortened version of one the blt files to test the parser but keep the others out the repo for now!

src/votekit/cvr_loaders.py
```diff
@@ -68,3 +69,89 @@ def rank_column_csv(
...
+                # first number is number of candidates, second is number of seats to elect
+                metadata = [int(data) for data in s.split(" ")]
+                if len(metadata) != 2:
+                    raise DataError(
+                        "metadata (first line) should have two parameters"
+                        " (number of candidates, number of seats)"
+                    )
+                seats = metadata[1]
+            # read in ballots, cleaning out rankings labeled '0' (designating end of line)
+            if numbers:
```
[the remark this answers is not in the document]
roryhe: not sure exactly what you mean by the first question, but I think in the current implementation it checks each line until it finds a 0 as the first word in a line, which indicates the end of the ballots (that is, the "numbers"). 'numbers' remains equal to True provided we haven't reached this point yet.

src/votekit/cvr_loaders.py
roryhe: I believe the first rstrip call strips the newline character, and the second strips any whitespace from the end of the line. (Jamie wrote this part of the code though, not me)

src/votekit/cvr_loaders.py
roryhe: ok, I think I removed it? but not sure if this got connected to this PR

tests/test_rand_transfer.py
```diff
@@ -0,0 +1,78 @@
...
+
+    rand_transfer_election = STV(pp, random_transfer, seats=seats)
+
+    while not rand_transfer_election.is_complete():
+        pp, out = rand_transfer_election.run_step(pp)
+
+    rand_transfer_outcome = out
+
+    # run election with fractional transfer and see if you get the same result
+    # (possible that this will fail sometimes but usually should succeed)
```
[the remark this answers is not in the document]
roryhe: not sure exactly what you mean (predicts the winner according to what (other) voting system?) Maybe I should just toss this one altogether?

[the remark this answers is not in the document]
jgibson517: Yeah, i think my point was that since we have tests for the STV class, a better unit test would be just to check the transfer function rather then check it in the context of STV

src/votekit/cvr_loaders.py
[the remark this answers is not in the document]
jgibson517: Got it! misunderstood the file structure

PR #33, opened 2023-07-21
Added outcome_model

MalavikaMukundan:


src/votekit/outcome_model.py
```diff
@@ -0,0 +1,134 @@
...
+    winners_votes (a Dict of candidates and list of their ballots):
+    each winner's list of ballots that elected them
+    previous: an instance of Outcome representing the previous round
+    """
+
+    curr_round: int
+    elected: list[str] = []
+    eliminated: list[str] = []
+    remaining: list[str] = []
+    profile: Optional[PreferenceProfile] = None
```
jgibson517, line 30: Since we're already returning profiles when we run elections, I don't think we need to store them in outcome class as well

jgibson517 (changes requested):
two small comments, and then this is good to go!

src/votekit/outcome_model.py
MalavikaMukundan: I think I disagree. Most often our workflow is gonna be geared towards accessing the outcome via the election, but sometimes, long after we've run elections and have a bunch of outcomes we might wanna do some analysis with the outcomes and be able to access the corresponding profiles. We're not storing the outcome as an attribute of the profile; we should at least have the option of storing the profile as an attribute of the outcome.

src/votekit/outcome_model.py
```diff
@@ -0,0 +1,128 @@
...
+    #     missing = (winners.difference(set(self.remaining)) |
+    # (losersdifference(set(self.remaining)))
+    #     raise ValueError(f"Cannot promote winners, {missing} not in remaining")
+    # return Outcome(
+    #     remaining=set(self.remaining).difference(winners | losers),
+    #     elected=list(set(self.elected) | winners)
+    #     eliminated=list(set(self.eliminated) | losers)
+    # )
+
+    def difference_remaining_candidates(
```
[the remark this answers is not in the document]
MalavikaMukundan: I commented it out in the final version of the file. Honestly this is such a specific thing to do that we can probably do it by hand when we need to without creating a function in the outcome file

src/votekit/outcome_model.py
MalavikaMukundan: I can however change the code so that we store profile only if it has not been stored in any previous round. That should take care of overuse of memory etc

PR #50, opened 2023-07-24
Small changes

MalavikaMukundan:
1. Largest changes are integrating state into election_types, one change to state where I make remaining non-optional again and deleted get_profile
2. Modified tests of both files above

### DIFF
--- a/src/votekit/election_state.py
+++ b/src/votekit/election_state.py
@@ -116,5 +116,5 @@ class ElectionState(BaseModel):
         changes = {}
         for index, candidate in curr_ranking.items():
             if prev_ranking[candidate] != index:
-                changes[candidate] = (prev_ranking[candidate}, index)
+                changes[candidate] = (prev_ranking[candidate], index)
         return changes
--- a/src/votekit/election_state.py
+++ b/src/votekit/election_state.py
@@ -3,7 +3,7 @@ from .profile import PreferenceProfile
 from typing import Optional
 
 
-# 1. is add_winners_and_losers necessary? --A: probabably not
+# 1. is add_winners_and_losers necessary? --A: probably not
 # 2. do we want anything more?]
 # 3. RCV cruncher: inspiration for functions--> statistics functions/things
 # Example of immutable data model for results
@@ -26,7 +26,7 @@ class ElectionState(BaseModel):
     curr_round: int
     elected: list[str] = []
     eliminated: list[str] = []
-    remaining: Optional[list] = None
+    remaining: list[str] = []
     profile: PreferenceProfile
     winner_votes: Optional[dict] = None
     previous: Optional["ElectionState"] = None
@@ -53,14 +53,8 @@ class ElectionState(BaseModel):
 
     def get_rankings(self) -> list[str]:
         """returns all candidates in order of their ranking at the end of the current round"""
-        if self.remaining:
-            return self.get_all_winners() + self.remaining + self.get_all_eliminated()
 
-        return self.get_all_winners() + self.get_all_eliminated()
-
-    def get_profile(self) -> PreferenceProfile:
-        """returns the election profile if it has been stored in any round up to the current one"""
-        return self.profile
+        return self.get_all_winners() + self.remaining + self.get_all_eliminated()
 
     def get_round_outcome(self, roundNum: int) -> dict:
         # {'elected':list[str], 'eliminated':list[str]}
@@ -114,15 +108,14 @@ class ElectionState(BaseModel):
         if not self.previous:
             raise ValueError("This is the first round, cannot compare previous ranking")
 
-        else:
-            prev_ranking = self.previous.get_rankings()
-            curr_ranking = self.get_rankings()
-            if curr_ranking == prev_ranking:
-                print("No changes in ranking")
+        prev_ranking = self.previous.get_rankings()
+        curr_ranking = self.get_rankings()
+        if curr_ranking == prev_ranking:
+            print("No changes in ranking")
 
-            changes = {}
-            for index, candidate in enumerate(curr_ranking):
-                if candidate != prev_ranking[index]:
-                    prev_rank = prev_ranking.index(candidate)
-                    changes[candidate] = (prev_rank, index)
-            return changes
+        changes = {}
+        for index, candidate in enumerate(curr_ranking):
+            if candidate != prev_ranking[index]:
+                prev_rank = prev_ranking.index(candidate)
+                changes[candidate] = (prev_rank, index)
+        return changes
--- a/src/votekit/election_state.py
+++ b/src/votekit/election_state.py
@@ -108,14 +108,13 @@ class ElectionState(BaseModel):
         if not self.previous:
             raise ValueError("This is the first round, cannot compare previous ranking")
 
-        prev_ranking = self.previous.get_rankings()
-        curr_ranking = self.get_rankings()
+        prev_ranking = {cand:index for index,cand in enumerate(self.previous.get_rankings())}
+        curr_ranking = {cand:index for index,cand in enumerate(self.get_rankings())}
         if curr_ranking == prev_ranking:
-            print("No changes in ranking")
+            return {}
 
         changes = {}
-        for index, candidate in enumerate(curr_ranking):
-            if candidate != prev_ranking[index]:
-                prev_rank = prev_ranking.index(candidate)
-                changes[candidate] = (prev_rank, index)
+        for index, candidate in curr_ranking.items():
+            if prev_ranking[candidate] != index:
+                changes[candidate] = (prev_ranking[candidate}, index)
         return changes
--- a/src/votekit/election_types.py
+++ b/src/votekit/election_types.py
@@ -1,98 +1,125 @@
 from .profile import PreferenceProfile
 from .ballot import Ballot
-from .models import Outcome
+from .election_state import ElectionState
 from typing import Callable
 import random
 from fractions import Fraction
 from copy import deepcopy
 
-
+##We're not passing winner_votes to outcome right now
+# If we want to track which candidate each ballot is ultimately going to, we need to do more.
+# TODO:
+# 1. winner_votes to election_state is supposed to have all the ballots in the
+# initial profile that ended up going to any elected candidate.
+# Bit tricky since all of our methods work with current state of profile not the initial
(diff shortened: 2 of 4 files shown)


src/votekit/election_types.py
```diff
@@ -1,98 +1,125 @@
...
+            for candidate in fp_order:
+                if fp_votes[candidate] >= self.threshold:
+                    elected.append(candidate)
+                    remaining.remove(candidate)
+                    ballots = self.transfer(
+                        candidate, ballots, fp_votes, self.threshold
+                    )
+        # since no one has crossed threshold, eliminate one of the people
+        # with least first place votes
+        else:
```
jgibson517, line 76: lets use if self.next_round(): here instead of else

src/votekit/election_types.py
```diff
@@ -1,98 +1,125 @@
...
+        fp_votes: dict = compute_votes(remaining, ballots)
+        fp_order = [
+            y[0] for y in sorted(fp_votes.items(), key=lambda x: x[1], reverse=True)
+        ]
+        elected = []
+        eliminated = []
+
+        # if number of remaining candidates equals number of remaining seats, everyone is elected
+        if len(remaining) == self.seats - len(self.election_state.get_all_winners()):
+            elected = fp_order
```
jgibson517, line 60: we should set self.election_state and return here because the election is over is this conditional is true

src/votekit/election_types.py
```diff
@@ -1,98 +1,125 @@
...
 import random
 from fractions import Fraction
 from copy import deepcopy
 
-
+##We're not passing winner_votes to outcome right now
+# If we want to track which candidate each ballot is ultimately going to, we need to do more.
+# TODO:
+# 1. winner_votes to election_state is supposed to have all the ballots in the
+# initial profile that ended up going to any elected candidate.
```
jgibson517, line 13: to this point i think we should store the initial profile in the STV class, just to have easy access to it. Also since the user has to pass a PreferenceProfile to initialize the class it seems weird to immediately store that in a separate object. It can be a hidden attribute if we don't want users calling it from the STV class
`
self.__profile = profile
`

src/votekit/election_types.py
```diff
@@ -1,98 +1,125 @@
...
-        self.eliminated: set = set()
-        self.seats = seats
-        self.threshold = self.get_threshold()
+        self.transfer: Callable = transfer
+        self.seats: int = seats
+
+        fp_votes = compute_votes(profile.get_candidates(), profile.get_ballots())
+        fp_order = [
+            y[0] for y in sorted(fp_votes.items(), key=lambda x: x[1], reverse=True)
+        ]
```
jgibson517, line 24: these lines get repeated a few times, lets turn them into a order_cands() method that we can call when initializing election state and during the actual running of STV
`
ElectionState(
            curr_round=0, elected=[], eliminated=[], remaining=self.order_cands(), profile=profile
        )
`

src/votekit/election_types.py
```diff
@@ -1,98 +1,125 @@
...
-            self.eliminated.add(lp_cand)
-
-        return PreferenceProfile(ballots=ballots), Outcome(
-            elected=self.elected,
-            eliminated=self.eliminated,
-            remaining=set(candidates),
-            votes=fp_votes,
+            eliminated.append(random.choice(lp_candidates))
+            ballots = remove_cand(eliminated[0], ballots)
+            remaining.remove(eliminated[0])
```
jgibson517, line 84: we should set `lp_cand = random.choice(lp_candidates)` as a variable so a) its clearer what is happening vs just indexing the list and b) I'm not sure thats the only way to break ties in this instance so it leaves room to abstract that

jgibson517 (changes requested):
a couple small suggestions, otherwise I think this looks good

src/votekit/election_types.py
MalavikaMukundan: I'm gonna use elif instead, because we don't wanna both elect and eliminate ppl in the same round.

src/votekit/election_types.py
MalavikaMukundan: Done

MalavikaMukundan:
@jamesturk I don't know how else to request a review but I'd be glad if you could take a look at these four files

src/votekit/election_state.py
```diff
@@ -114,15 +108,14 @@ def changed_rankings(self) -> Optional[dict]:
...
-            changes = {}
-            for index, candidate in enumerate(curr_ranking):
-                if candidate != prev_ranking[index]:
-                    prev_rank = prev_ranking.index(candidate)
-                    changes[candidate] = (prev_rank, index)
-            return changes
+        prev_ranking = self.previous.get_rankings()
+        curr_ranking = self.get_rankings()
+        if curr_ranking == prev_ranking:
+            print("No changes in ranking")
```
jamesturk, line 114: generally speaking print statements like this don't belong in functions since they'll clutter up downstream output

if this is an error condition, consider raising an exception, or maybe you just want to return [] here to avoid executing the remaining code

src/votekit/election_state.py
```diff
@@ -114,15 +108,14 @@ def changed_rankings(self) -> Optional[dict]:
...
+
+        changes = {}
+        for index, candidate in enumerate(curr_ranking):
```
jamesturk, line 117: this is fine if you're pretty confident number of candidates will remain fairly small let's say N<100, if there's any reason to think it'll get much bigger than that, consider converting both curr/prev to lists of index,candidate so you don't need to do expensive prev.index() lookups as often

src/votekit/election_types.py
```diff
@@ -1,117 +1,140 @@
...
+# 2. integrate Cincinatti transfer
 class STV:
     def __init__(self, profile: PreferenceProfile, transfer: Callable, seats: int):
-        self.profile = profile
-        self.transfer = transfer
-        self.elected: set = set()
-        self.eliminated: set = set()
-        self.seats = seats
-        self.threshold = self.get_threshold()
+        self.__profile = profile
```
jamesturk, line 18: Q: is this double underscore intentional?

src/votekit/election_types.py
```diff
@@ -1,117 +1,140 @@
...
+        self.transfer: Callable = transfer
+        self.seats: int = seats
+        self.election_state: ElectionState = ElectionState(
```
jamesturk, line 21: note: you don't have to annotate these since the type can be inferred by the RHS of the equals

src/votekit/election_types.py
```diff
@@ -1,117 +1,140 @@
...
+            curr_round=0,
+            elected=[],
+            eliminated=[],
+            remaining=compute_votes(profile.get_candidates(), profile.get_ballots())[1],
```
jamesturk, line 25: I'd add a comment here to explain the [1]

src/votekit/election_types.py
```diff
@@ -1,117 +1,140 @@
...
+            profile=profile,
+        )
+        self.threshold: float = self.get_threshold()
```
jamesturk, line 28: annotated as float, but get_threshold returns an int

src/votekit/election_types.py
```diff
@@ -1,117 +1,140 @@
...
-            if fp_votes[candidate] >= self.threshold:
-                self.elected.add(candidate)
-                candidates.remove(candidate)
-                ballots = self.transfer(candidate, ballots, fp_votes, self.threshold)
-
-        if self.next_round():
+        ##TODO:must change the way we pass winner_votes
+        remaining: list[str] = self.election_state.remaining
+        ballots: list[Ballot] = self.election_state.profile.get_ballots()
+        fp_votes: dict
```
jamesturk, line 50: some variables declared and not used here

src/votekit/election_types.py
```diff
@@ -1,117 +1,140 @@
...
             )
 
         while self.next_round():
-            profile, outcome = self.run_step(profile)
+            self.run_step()
+
+        return self.election_state
 
-        return outcome
+    def get_init_profile(self):
```
jamesturk, line 109: if the profile is accessible, any reason to use this function instead of accessing it as self.profile? (or self.initial_profile if you want to be more clear about its purpose)

src/votekit/election_types.py
```diff
@@ -1,117 +1,140 @@
...
-
     for candidate in candidates:
         weight = Fraction(0)
         for ballot in ballots:
             if ballot.ranking and ballot.ranking[0] == {candidate}:
                 weight += ballot.weight
         votes[candidate] = weight
 
-    return votes
+    order = [y[0] for y in sorted(votes.items(), key=lambda x: x[1], reverse=True)]
```
jamesturk, line 129: I wonder if this is the best way to do this, effectively returning the elements twice.  Might consider creating a new dictionary with correct order, or return a list with (k, v) tuples, instead of both.

jamesturk (commented):
Looking good. Left some comments on what is there, looks like some tests aren't implemented yet either.

src/votekit/election_state.py
MalavikaMukundan: Ok I'm changing it to return {} for now since we'd otherwise return a dictionary

src/votekit/election_types.py
MalavikaMukundan: Yes, Jack said to make this profile a hidden attribute because we don't want the user to accidentally mess with it

src/votekit/election_types.py
```diff
@@ -1,117 +1,140 @@
...
-                candidates.remove(candidate)
-                ballots = self.transfer(candidate, ballots, fp_votes, self.threshold)
-
-        if self.next_round():
+        ##TODO:must change the way we pass winner_votes
+        remaining: list[str] = self.election_state.remaining
+        ballots: list[Ballot] = self.election_state.profile.get_ballots()
+        fp_votes: dict
+        fp_order: list[str]
+        fp_votes, fp_order = compute_votes(remaining, ballots)
```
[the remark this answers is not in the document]
MalavikaMukundan: first place, added as comment

src/votekit/election_types.py
MalavikaMukundan: Ok I've changed this by just returning the list and eliminating the use of the fp_order list everywhere

MalavikaMukundan:
Ok this is failing some linting now and I'm gonna close it and repoen a new PR

PR #51, opened 2023-07-25
Undervote check

edelanuez:
this is the undervote code that takes in a ballot with empty candidate spaces and outputs a truncated ballot


src/votekit/cleaning.py
```diff
@@ -173,3 +172,27 @@ def remove_from_ballots(ballot: Ballot, non_cands: list[str]) -> Ballot:
...
+    Args:
+        ballot (Ballot): a ballot with empty ranks in a particular ranking list
+
+    Returns:
+        Ballot: a ballot without emptinesss
+    """
+
+    def undervote(ballot: Ballot) -> Ballot:
+        rank_list = ballot.ranking
+        cleaned_rank_list = [rank for rank in rank_list if " " not in rank]
```
jgibson517, line 189: the parser defaults empty cells (aka undervotes) as `None` so lets replace `'' ''` with `None`

jgibson517 (changes requested):
One small tweak and then this is good!

src/votekit/cleaning.py
edelanuez: got it

src/votekit/metrics.py
```diff
@@ -0,0 +1,146 @@
...
+    n = len(list1)
+    indices = list(range(n))
+    model_dictionary = dict(zip(list1, indices))
+    target_dictionary = {key: model_dictionary[key] for key in list2}
+    unranked_values = list(target_dictionary.values())
+    return unranked_values
+
+
+def perform_bubble_sort(
+    unorderedlist,
```
jgibson517, line 35: why is this an underorderedlist vs a regular list?

src/votekit/metrics.py
```diff
@@ -0,0 +1,146 @@
...
+):  # compare the intial ranking to whatever order the candidates are listed as
+    n = len(list1)
+    indices = list(range(n))
+    model_dictionary = dict(zip(list1, indices))
+    target_dictionary = {key: model_dictionary[key] for key in list2}
+    unranked_values = list(target_dictionary.values())
+    return unranked_values
+
+
+def perform_bubble_sort(
```
jgibson517, line 34: lets call this kendall_tau instead of bubble sort

src/votekit/metrics.py
```diff
@@ -0,0 +1,146 @@
+from votekit.cvr_loaders import rank_column_csv
```
jgibson517, line 1: what is the rest of the code in this file, is it just scratch work/testing materials?

jgibson517 (changes requested):

src/votekit/metrics.py
```diff
@@ -0,0 +1,146 @@
...
+    model_dictionary = dict(zip(list1, indices))
+    target_dictionary = {key: model_dictionary[key] for key in list2}
+    unranked_values = list(target_dictionary.values())
+    return unranked_values
+
+
+def perform_bubble_sort(
+    unorderedlist,
+):  # do the actual bubble sort and get the distance
+
```
jgibson517, line 37: this function should also have a doc string explaining what it does and inputs/outputs

jgibson517 (changes requested):

poetry.lock
jamesturk, line 1: `poetry.lock` can be removed from this repository (and added to .gitignore) since it is a library, `pyproject.toml`  is all you'll need for distributing this & it will save everyone's commits from having spurious lockfile updates

src/votekit/cleaning.py
```diff
@@ -173,3 +172,27 @@ def remove_from_ballots(ballot: Ballot, non_cands: list[str]) -> Ballot:
...
+    """
+    takes a ballot and truncates its rankings in the case.
+    Args:
+        ballot (Ballot): a ballot with empty slots in a particular ranking list
+
+    Returns:
+        Ballot: a truncated ballot without empty slots
+    """
+
+    def undervote(ballot: Ballot) -> Ballot:
```
jamesturk, line 187: this should not be a closure (inner function) since it does not use any of the containing scope

src/votekit/metrics.py
```diff
@@ -0,0 +1,95 @@
+from votekit.cvr_loaders import rank_column_csv
+from votekit.election_types import STV, fractional_transfer
+
+
+fpath = "/Users/emariedelanuez/VoteKit/src/votekit/mn_clean_ballots.csv"
```
jamesturk, line 5: there should not be hard-coded paths, or global variables like this in the library

src/votekit/metrics.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+def loadin(fpath):
```
jamesturk, line 8: this function doesn't do anything

src/votekit/metrics.py
```diff
@@ -0,0 +1,95 @@
...
+    The goal of this file is to output the distance
+    between the rankings produced by the STV
+    results of a race and an unordered
+    list of candidates.
+    """
+    data = rank_column_csv(fpath)
+    return data
+
+
+data = loadin(fpath)
```
jamesturk, line 19: global variables like this should be removed (especially ones w/ names that mask parameters like `data`)

src/votekit/metrics.py
```diff
@@ -0,0 +1,95 @@
...
+            if unorderedlist[index_j - 1] > unorderedlist[index_j]:
+                swapcount += 1
+                unorderedlist[index_j - 1], unorderedlist[index_j] = (
+                    unorderedlist[index_j],
+                    unorderedlist[index_j - 1],
+                )
+    return swapcount
+
+
+rankinglist, candidates = stv_mini(data)
```
jamesturk, line 92: this code belongs in a test file or similar, there should be no code at the top level of a file intended to be used as a library

src/votekit/metrics.py
```diff
@@ -0,0 +1,95 @@
...
+    which assigns indices to the ranked candidate list
+    through dictionary comprehension.
+    Then, we create a target dictionary, where the keys are candidates in the unordered list,
+    and assigns its corresponding values from the model dictionary.
+    Then, we make a list of the values in our target dictionary,
+    which preserves the order of the "unranking."\n
+    """
+    n = len(list1)
+    indices = list(range(n))
+    model_dictionary = dict(zip(list1, indices))
```
jamesturk, line 64: this can be written much more efficiently using `enumerate` https://realpython.com/python-enumerate/

src/votekit/metrics.py
```diff
@@ -0,0 +1,95 @@
...
+def kendall_tau(unorderedlist):
+    """ "
+    kendall_tau:
+    takes in: the unordered, but indexed candidate list from the comparision function
+    returns: swap distance\n
+    the function does the bubble sort on the indices of the unordered candidate list
+    and returns a swap distance value.\n
+    """
+    swapcount = 0
+    for index_i in range(len(unorderedlist)):
```
jamesturk, line 79: see above note about `enumerate`

jamesturk (changes requested):

src/votekit/cleaning.py
edelanuez: by no inner function, do you mean just left tab the undervote function and returning a ballot?  something like this (sorry if this is ugly to read):
# def undervote_profile(pp: PreferenceProfile) -> PreferenceProfile:
#     """
#     takes a ballot and truncates its rankings in the case.
#     Args:
#         ballot (Ballot): a ballot with empty slots in a particular ranking list

#     Returns:
#         Ballot: a truncated ballot without empty slots
#     """

def undervote(ballot: Ballot) -> Ballot:
    rank_list = ballot.ranking
    cleaned_rank_list = [rank for rank in rank_list if None not in rank]
    return Ballot(
        id=ballot.id,
        ranking=cleaned_rank_list,
        weight=Fraction(ballot.weight),
        voters=ballot.voters,
    )

    # pp_clean = _clean(pp=pp, clean_ballot_func=undervote)
    # return pp_clean

jamesturk: Try editing your comment to use ``` around code, as is it got pretty mangled.

cdonnay:
Closing since we have drifted so far from votekit 0.1

PR #54, opened 2023-07-27
Election functions complete w tests

ziglaser:
Added election functions for Limited, Bloc, SNTV, SNTV-STV Hybrid, and Top Two, along with a testing file


src/votekit/pref_profile.py
```diff
@@ -1,4 +1,4 @@
-from .ballot import Ballot
+from ballot import Ballot
```
jamesturk, line 1: most likely you're executing the file directly which is causing it to need this import to be written this way but this will break imports from within the package, since the top level package is named votekit it won't be able to find ballot without the relative path (.)

tests/test_election_types2.py
```diff
@@ -0,0 +1,139 @@
+import votekit.election_types as et
+from votekit.pref_profile import PreferenceProfile
+from votekit.ballot import Ballot
+from votekit.election_state import ElectionState
+from fractions import Fraction
+
+
+def list_equal(list1, list2):
```
jamesturk, line 8: this could be list1 == list2

tests/test_election_types2.py
```diff
@@ -0,0 +1,139 @@
...
+        r1_cutoff=r1_cutoff,
+        seats=seats,
+        transfer=et.fractional_transfer,
+    )
+    outcome = hybrid_election.run_election()
+    return equal_ElectionStates(outcome, target_state)
+
+
+def test_battery_Bloc():
+    test_profile = read_profile("data/zach_test_ballots.csv")
```
jamesturk, line 59: one thing to consider: if future code breaks these tests, will a person looking a the test be able to understand the likely culprit and where to look?  I'd consider either adding some comments to explain what zach_test_ballots looks like and what expected results would be and why, or building smaller, more explicit tests that test specific behaviors of the various methods

tests/test_election_types2.py
```diff
@@ -0,0 +1,139 @@
...
+    sntv_target2 = ElectionState(
+        curr_round=1,
+        elected=["B", "A", "C", "D", "G"],
+        eliminated=["H", "F", "E", "I"],
+        remaining=[],
+        profile=PreferenceProfile(),
+    )
+    sntv_test2 = _test_SNTV(profile=test_profile, seats=5, target_state=sntv_target2)
+
+    assert sntv_test1 and sntv_test2
```
jamesturk, line 112: since you're testing two things, having the test fail early if sntv_test1 fails (and similar in other functions) would probably be preferable for debugging down the line

instead of asserting var1 and var2, you could assert the result of each _test_ call as it is made for failures that are more easily linked to the issue at hand

jgibson517 (changes requested):

ziglaser:
@jamesturk @jgibson517 PR is now updated with end of week objectives. Here are some assorted comments:
- Added graphs.py file which contains PairwiseComparisonGraph object. This will likely get folded into a file with the ballot graph, or share a subfolder with it. Added tests for all methods in test_graphs.py.
- Added metrics.py file which contains only borda_scores, a function that computes a dictionary of borda scores for an election. This belongs in a class of general utility methods to get data about any preference profile we want  (ie first place votes, borda scores, etc). It will probably wind up with a different name under the metrics subfolder. Added tests_metrics.py.
- Added election systems for DominatingSets and CondoBorda (using PWC graph) into election_types.py, along with new tests for those in election_tests_2.py
- Slightly modified election_state.py code so that it doesn't require results be lists of strings, it now allow lists of, well, anything, but relevantly sets which will track ties in the outcome (we should likely fully transition outcomes to be sets to allow ties and mirror Ballot at some point)

src/votekit/profile.py
```diff
@@ -12,9 +12,16 @@ class PreferenceProfile(BaseModel):
     candidates (list): list of candidates, can be user defined
     """
 
-    ballots: list[Ballot]
+    ballots: list[Ballot] = list()
     candidates: Optional[list] = None
 
+    def __eq__(self, other):
```
jamesturk, line 18: this __eq__ method is pretty forgiving, PreferenceProfiles are equal to all non-PP objects and the only thing that's checked is that all ballots are in the other PP, not vice versa.  this could lead to some confusing behavior down the line

maybe add a few tests for equal/unequal PPs to flesh this out more

src/votekit/graphs.py
```diff
@@ -0,0 +1,162 @@
...
+                if cand1 in s:
+                    count += ballot.weight
+                    break
+                elif cand2 in s:
+                    break
+        return Fraction(count)
+
+    def compute_pairwise_dict(self) -> dict:
+        pairwise_dict = {}  # {(cand_a, cand_b): freq cand_a is preferred over cand_b}
+        cand_pairs = list(combinations(self.candidates, 2))
```
jamesturk, line 60: tip: less memory intensive to not cast this to list

src/votekit/election_types.py
```diff
@@ -139,3 +408,20 @@ def remove_cand(removed_cand: str, ballots: list[Ballot]) -> list[Ballot]:
         update[n].ranking = new_ranking
 
     return update
+
+
+def remove_cand_set(removed_set: Iterable[str], ballots: list[Ballot]) -> list[Ballot]:
```
jamesturk, line 413: because remove_cand does a deepcopy this method winds up doing a ton of intermediate copies, might be worth altering to either have remove_cand take an iterable[str] or write this without relying on remove_cand and just do your own single deep copy at the start

src/votekit/graphs.py
```diff
@@ -0,0 +1,162 @@
...
+            arrows=True,
+            alpha=1,
+            node_size=1000,
+            arrowsize=25,
+        )
+        edge_labels = {(i, j): G[i][j]["weight"] for i, j in G.edges()}
+        nx.draw_networkx_edge_labels(
+            G, pos, edge_labels=edge_labels, label_pos=0.5, font_size=10
+        )
+        # Out stuff
```
jamesturk, line 110: tip: since you always want one of these to be true, it's worth considering combining them into one parameter

perhaps the default behavior is show=True, and iff outfile is passed that is used instead, otherwise show=False, outfile=None means this does nothing

src/votekit/graphs.py
```diff
@@ -0,0 +1,162 @@
...
+        for k, v in beat_set_size_dict.items():
+            if v in tier_dict.keys():
+                tier_dict[v].add(k)
+            else:
+                tier_dict[v] = {k}
+        tier_list = [tier_dict[k] for k in sorted(tier_dict.keys(), reverse=True)]
+        return tier_list
+
+
+if __name__ == "__main__":
```
jamesturk, line 145: maybe a test?

tests/test_elections_2.py
```diff
@@ -0,0 +1,185 @@
...
+    ),
+]
+CONDO_TEST_PROFILE = PreferenceProfile(ballots=condo_ballot_list)
+
+
+def equal_electionstates(state1, state2):
+    print(state1.get_all_winners(), state1.get_all_eliminated(), state1.get_rankings())
+    print(state2.get_all_winners(), state2.get_all_eliminated(), state2.get_rankings())
+
+    return (
```
jamesturk, line 32: output for errors is pretty tough to debug, probably why you have the print statements in here

I'd recommend instead using multiple assert statements, `pytest` will highlight the errors better if you assert winners1=winners2 then assert elim1 == elim2, etc.

jamesturk (changes requested):

ziglaser:
@jamesturk @jgibson517 @realdavemccune Made changes in line with James' comments, the most extensive of which are a merge and rewrite of the remove_cand/remove_cand_sets functions, a rewrite of the PreferenceProfile __eq__ function, along with corresponding tests, and the addition of a Ballot __eq__ function

PR #56, opened 2023-07-28
Chamberlain-Courant, Approval, Range Types

andrewmhong:
Added Chamberlain-Courant, Approval, Range election type classes; updated PreferenceProfile class (in profile.py) with new RangeBallot class (in ballot.py)

### DIFF
--- a/src/votekit/ballot.py
+++ b/src/votekit/ballot.py
@@ -18,3 +18,20 @@ class Ballot(BaseModel):
 
     class Config:
         arbitrary_types_allowed = True
+
+        
+class RangeBallot(BaseModel):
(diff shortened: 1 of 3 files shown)


realdavemccune:
I was looking at the Cham Cour code, and sorry if I missed it but how are partial ballots being processed? Or does this function assume complete ballots? Also, does the function handle ties? Ties are obnoxious so I would deal with them in the easiest possible way and just document the choice.

andrewmhong:
>

cdonnay:
Election class revamp from 3.0.0 means this will need to be restarted if we want to include it.

PR #61, opened 2023-08-04
updates to blt/random transfer/quota PR based on Jack's comments

roryhe:


tests/data/mn_2021_Ward_2.csv
jgibson517, line 1: Any reason why this file is included? It doesn't seem to be referenced in your tests? Otherwise I think this is good to go!

jgibson517 (changes requested):

tests/data/mn_2021_Ward_2.csv
roryhe: Oh no, that can go. Are you able to remove it?

PR #65, opened 2023-08-08
Adding Sequential RCV to elections types

bmacias01:
@jamesturk  Here is some code to be reviewed for a sequential RCV election.


bmacias01:
@realdavemccune Here is the PR for seqRCV

bmacias01:
@jamesturk here are the commits for added distance functions to compare preference profiles. Earth mover and Lp distance

realdavemccune:
This looks good. As discussed, should probably add something how dealing with ties, or at least document how ties are handled (currently, handled from STV function).

metrics/distances.py
```diff
@@ -0,0 +1,86 @@
+from votekit.profile import PreferenceProfile
+import numpy as np
+import ot
```
jamesturk, line 3: make sure whatever library this is gets added to pyproject.toml

metrics/distances.py
```diff
@@ -0,0 +1,86 @@
...
+    # Floyd Warshall Shortest Distance alorithm. Returns a dictionary of shortest path for each node
+    FW_dist_dict = nx.floyd_warshall(ballot_graph)
+    keysList = list(FW_dist_dict.keys())
+    keysList.sort()
+    cost_matrix = np.zeros((len(keysList), len(keysList)))
+    for i in range(len(keysList)):
+        node_dict = FW_dist_dict[keysList[i]]
+        cost_col = [value for key, value in sorted(node_dict.items())]
+        cost_matrix[i] = cost_col
+    G0 = ot.emd(electA_distr, electB_distr, cost_matrix)
```
jamesturk, line 34: tip: prefer descriptive variable names, what is G0 (also capitalization is usually reserved for constants)

metrics/distances.py
```diff
@@ -0,0 +1,86 @@
...
+    # Solving Earth Mover Distance
+    election_arrays = profilePairs_to_arrays(pp1, pp2)
+    electA_distr = election_arrays[0]
+    electB_distr = election_arrays[1]
+
+    # TODO:create ballot graph:
+
+    # Floyd Warshall Shortest Distance alorithm. Returns a dictionary of shortest path for each node
+    FW_dist_dict = nx.floyd_warshall(ballot_graph)
+    keysList = list(FW_dist_dict.keys())
```
jamesturk, line 27: can be shortened to `keys_list = sorted(fw_dict.keys())`

metrics/distances.py
```diff
@@ -0,0 +1,86 @@
...
+        cost_col = [value for key, value in sorted(node_dict.items())]
+        cost_matrix[i] = cost_col
+    G0 = ot.emd(electA_distr, electB_distr, cost_matrix)
+
+    # Hadamard Product = Earth mover dist between two matrices
+    earth_mover_dist = np.sum(np.multiply(cost_matrix, G0))
+    return earth_mover_dist
+
+
+def Lp_dist(
```
jamesturk, line 41: tip: public functions should be lower case and have meaningful names

src/votekit/election_types.py
```diff
@@ -333,6 +395,17 @@ def random_transfer(
...
 
+def seqRCV_transfer(
+    winner: str, ballots: list[Ballot], votes: dict, threshold: int
+) -> list[Ballot]:
+    """
+    Useful for a Sequential RCV election which does not use a transfer method ballots \n
+    ballots: list of ballots \n
+    output: same ballot list
+    """
+    return ballots
```
jamesturk, line 406: a comment might be useful to explain the use of this since it looks "wrong" at first glance, but has reasons to exist

src/votekit/profile.py
```diff
@@ -56,17 +54,21 @@ def num_ballots(self) -> Fraction:
...
+        input: standardize to model an election distribution
         """
+        num_ballots = self.num_ballots()
         di: dict = {}
         for ballot in self.ballots:
             if str(ballot.ranking) not in di.keys():
                 di[str(ballot.ranking)] = Fraction(0)
-            di[str(ballot.ranking)] += ballot.weight
-
+            if standardize is True:
```
jamesturk, line 68: tip: can drop `is True`

src/votekit/profile.py
```diff
@@ -56,17 +54,21 @@ def num_ballots(self) -> Fraction:
...
+                di[str(ballot.ranking)] += ballot.weight / num_ballots
```
jamesturk, line 69: tip can use self.num_ballots() here instead of computing the variable when it isn't needed

visuals/MDS.py
jamesturk, line 1: always name Python files meant to be imported by others w/ lower case-- case sensitivity differences on file systems cause problems otherwise (Windows/MacOS/Linux all treat this differently)

visuals/MDS.py
```diff
@@ -0,0 +1,74 @@
+from votekit.profile import PreferenceProfile
+from metrics.distances import earth_mover_dist, Lp_dist
+import matplotlib.pyplot as plt
+import numpy as np
+from sklearn import manifold
+import sys
+
+sys.path.append("../")
```
jamesturk, line 8: this should not be in this file, importing a module should not change `sys.path` since a lot of weird things can happen if you do

visuals/MDS.py
```diff
@@ -0,0 +1,74 @@
...
+    pp_arr2 = list(pp_set2)
+
+    if distance == "L1":
+        for i in range(rows):
+            for j in range(cols):
+                if i != j:
+                    dist_matrix[i][j] = Lp_dist(pp_arr1[i], pp_arr2[j], 1)
+        return dist_matrix
+
+    if distance == "earth mover":
```
jamesturk, line 37: should be `elif` as it is a continuation of above

visuals/MDS.py
```diff
@@ -0,0 +1,74 @@
...
+    pp_set1 and pp_set2: set of preference profiles \n
+    distance: Use "L1" for L1 dist, "earth mover" for earth mover dist.
+    """
+    rows = len(pp_set1)
+    cols = len(pp_set2)
+    dist_matrix = np.zeros((rows, cols))
+    pp_arr1 = list(pp_set1)
+    pp_arr2 = list(pp_set2)
+
+    if distance == "L1":
```
jamesturk, line 30: you might consider instead passing in a function (either Lp_dist or eearth_mover_dist) that would make this extensible to other distance measures without requiring string names of each one as you go

jamesturk (changes requested):

metrics/distances.py
bmacias01: done

metrics/distances.py
bmacias01: done

metrics/distances.py
bmacias01: done

src/votekit/profile.py
bmacias01: I did that above when num_ballots = self.num_ballots().

visuals/MDS.py
bmacias01: done

visuals/MDS.py
bmacias01: done

metrics/distances.py
bmacias01: done @jamesturk , do I commit the pyproject.toml ?

src/votekit/profile.py
bmacias01: done

visuals/MDS.py
bmacias01: done

PR #66, opened 2023-08-09
Ballot Generation

jennjwang:
Need more tests, but creating an initial PR


tests/test_ballot_generator.py
```diff
@@ -0,0 +1,402 @@
...
+        bloc_voter_prop=bloc_voter_prop,
+    )
+
+    # Test
+    assert do_ballot_probs_match_ballot_dist(ballot_prob_dict, generated_profile)
+
+
+def test_BT_probability_calculation():
+
+    # Set-up
```
jamesturk, line 296: tip: this setup that gets copied and pasted could be refactored into a pytest.fixture: https://docs.pytest.org/en/7.1.x/how-to/fixtures.html (not necessary, but might be nice)

src/votekit/cvr_loaders.py
```diff
@@ -8,6 +8,8 @@
 from pandas.errors import EmptyDataError, DataError
 from fractions import Fraction
 
+# TODO: update documentation for function below
```
jamesturk, line 11: TODO

src/votekit/ballot_generator.py
```diff
@@ -0,0 +1,558 @@
...
+        super().__init__(**data)
+
+        # Assign additional parameters specific to AC
+        self.slate_to_candidate = slate_to_candidate
+        self.pref_interval_by_bloc = pref_interval_by_bloc
+        self.bloc_voter_prop = bloc_voter_prop
+        self.bloc_crossover_rate = bloc_crossover_rate
+
+        # assert sum(bloc_crossover_rate.values()) == 1
+        assert sum(bloc_voter_prop.values()) == 1
```
jamesturk, line 287: note: assertions can be disabled in production (https://realpython.com/python-assert-statement/#using-assert-for-data-processing-and-validation)

probably best to write something like:

```
if sum(...) != 1:
    raise ValueError("message here")
```

src/votekit/ballot_generator.py
```diff
@@ -0,0 +1,558 @@
...
+        Args:
+            pref_interval_by_bloc (dict): a mapping of slate to preference interval
+            (ex. {race: {candidate : interval length}})
+            bloc_voter_prop (dict): a mapping of slate to voter proportions
+            (ex. {race: voter proportion})
+        """
+        # Call the parent class's __init__ method to handle common parameters
+        super().__init__(**data)
+
+        assert sum(bloc_voter_prop.values()) == 1
```
jamesturk, line 186: see comment elsewhere about assert vs raise

jamesturk (commented):
left a few comments/suggestions but looks good overall!

PR #67, opened 2023-08-11
Hyperparams and Simulation base class

jgibson517:
- Edits to ballot generator base class to allow the input of hyperparameters (cohesion, dirichlet values, bloc sizes) 
- Base class with method for generated multiple sets of ballots at once for large simulations


tests/test_models.py
```diff
@@ -1,28 +1,50 @@
...
-        orig.add_winners_and_losers({"D"}, {"E"})
+    def sim_election():
+        pass
 
-    # ensure error message mentions both bad candidates
-    assert "D" in str(exc.value)
-    assert "E" in str(exc.value)
 
+class DummyActual(Simulation):
+    ballots = MagicMock(spec=PreferenceProfile)
```
jamesturk, line 19: 👍

src/votekit/models.py
```diff
@@ -1,34 +1,57 @@
...
-        if not winners.issubset(self.remaining) or not losers.issubset(self.remaining):
-            missing = (winners - self.remaining) | (losers - self.remaining)
-            raise ValueError(f"Cannot promote winners, {missing} not in remaining")
-        return Outcome(
-            remaining=self.remaining - winners - losers,
-            elected=self.elected | winners,
-            eliminated=self.eliminated | losers,
-        )
+    def __init__(self, ballots: Union[PreferenceProfile, dict, None] = None):
+        if ballots:
```
jamesturk, line 19: it might be worth assigning self.ballots even if not true, so that the attribute at least exists (a missing attribute error down the line would likely be surprising)

src/votekit/ballot_generator.py
```diff
@@ -0,0 +1,534 @@
...
+            prob_ballot_given_opp_first = {
+                ballot: freq / bloc_first_count
+                for ballot, freq in ballot_frequencies.items()
+                if ballot[0] == opp_bloc
+            }
+
+            bloc_voters = int(self.slate_voter_prop[bloc] * self.number_of_ballots)
+            pref_interval_dict = self.pref_interval_by_slate[bloc]
+            for _ in range(bloc_voters):
+                first_choice = np.random.choice(
```
jamesturk, line 461: I believe (at a cursory glance at least) this code could be simplified to instead choose between `prob_ballot_given_block_first` and `prob_ballot_given_opp_first` and then the `bloc_ordering = random.choices` call could just use the result of that weighted choice. (removing duplication in lines 468-480)

taking that even further, it doesn't look like both are used, it will always be one or the other.  you could defer the calculation itself, so something like:
```
# make weighted choice between blocs
# calculate the appropriate `prob_ballot`
# select `bloc_ordering` from `prob_ballot`
```

src/votekit/ballot_generator.py
```diff
@@ -0,0 +1,534 @@
...
+
+        pp = self.ballot_pool_to_profile(
+            ballot_pool=ballot_pool, candidates=self.candidates
+        )
+        return pp
+
+
+class BradleyTerry(BallotGenerator):
+    def __init__(
+        self, pref_interval_by_slate: dict = {}, slate_voter_prop: dict = {}, **data
```
jamesturk, line 223: mutable default arguments (see other note)

src/votekit/ballot_generator.py
```diff
@@ -0,0 +1,534 @@
...
+    @abstractmethod
+    def generate_profile(self) -> PreferenceProfile:
+        pass
+
+    # @staticmethod
+    # def list_to_set(candidates):
+    #     return [set(cand) for cand in candidates]
+
+    @staticmethod
+    def ballot_pool_to_profile(ballot_pool, candidates):
```
jamesturk, line 74: just a note on "pythonic", especially since the rest of your code has excellent style overall. 

Most people don't use `staticmethod`, when a function works as a staticmethod (as opposed to `classmethod`, which is useful!) it takes no parameters relevant to the class (i.e. no `self`, no `cls`) and is best served as a module-level utility method.

That said, this is a 100% fine way to do things, but unless you're intentionally leaving the window open for a class to override it, I'd say 9/10 python codebases would just move this to the module-level instead of class-level.

jamesturk (commented):
Looks great overall

PR #73, opened 2023-08-17
Updated Code Addressing James comments and adding MDS plotting code

bmacias01:


pyproject.toml
```diff
@@ -11,8 +11,8 @@ pandas = "^2.0.2"
 mypy = "^1.4.1"
 pydantic = "^1.10.9"
 pandas-stubs = "^2.0.2.230605"
```
jamesturk, line 13: Two issues here, first is you don't want to remove networkx/distinctipy, second is `sklearn` is https://pypi.org/project/scikit-learn/ and should be installed as `scikit-learn`.

jamesturk (commented):
a quick review looks good to me with the exception of a pyproject.toml issue, someone might want to give the actual code a second pass if they have time

pyproject.toml
bmacias01: ok should I readd those then do a another pull request?

jgibson517: I took care of it Brenda! it was an easy fix

PR #85, opened 2023-08-21
Adding MDS code and distance functions 

bmacias01:


src/votekit/metrics/distances.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+def earth_mover_dist(pp1: PreferenceProfile, pp2: PreferenceProfile) -> int:
+    """
+    Computes the earth mover distance between two elections. \n
+    Assumes both elections share the same candidates
+    """
+    # create ballot graph
+    graph = BallotGraph(source=pp2, complete=True)
+    ballot_graph = graph.from_profile(profile=pp2, complete=True)
```
jgibson517, line 16: if the source for the Graph is a preference profile it updates the weights, so this line is redundant

src/votekit/metrics/distances.py
```diff
@@ -0,0 +1,108 @@
...
+        diff = [x - y for x, y in zip(electA_distr, electB_distr)]
+        return max(diff)
+
+    else:
+        raise ValueError("Unsupported input type")
+
+
+# helper functions
+# these functions comvert a list of preference profiles into distribution arrays
+def profilePairs_to_arrays(
```
jgibson517, line 69: this is a super useful function!! but instead of limiting it to pairs, lets either allow users to input a list profiles or change it to profile_to_array() (i.e. single profile) and loop through when we use, bit more abstract that way!

src/votekit/metrics/distances.py
```diff
@@ -0,0 +1,108 @@
...
+
+    elect1 = combined_dict | elect1
+    elect2 = combined_dict | elect2
+
+    electA_distr = [float(elect1[key]) for key in sorted(elect1.keys())]
+    electB_distr = [float(elect2[key]) for key in sorted(elect2.keys())]
+    return electA_distr, electB_distr
+
+
+def em_array(pp: PreferenceProfile):
```
jgibson517, line 90: lets write out what em is, the code looks great I just don't know what this function is doing at first glance!

src/votekit/profile.py
```diff
@@ -52,17 +52,28 @@ def num_ballots(self):
...
+                if standardize:
+                    di[rank_tuple] = ballot.weight / num_ballots
+                else:
+                    di[rank_tuple] = ballot.weight
+
+            else:
+                if standardize:
+                    di[rank_tuple] += ballot.weight / num_ballots
+                else:
+                    di[rank_tuple] += ballot.weight
```
jgibson517, line 76:
```suggestion
            weight = ballot.weight
            if standardize:
                weight = ballot.weight / num_ballots
            if rank_tuple not in di.keys():
                di[rank_tuple] = weight
            di[rank_tuple] += weight
```

bmacias01: done

src/votekit/metrics/distances.py
bmacias01: done

src/votekit/metrics/distances.py
bmacias01: done

PR #90, opened 2023-08-25
Generalized IC/IAC - Ballot Simplex

jennjwang:


src/votekit/ballot_generator.py
```diff
@@ -39,32 +41,89 @@ def __init__(
...
+                if round(sum(interval.values())) != 1:
+                    raise ValueError("Preference interval for candidates must sum to 1")
+            if bloc_voter_prop.keys() != pref_interval_by_bloc.keys():
+                raise ValueError("slates and blocs are not the same")
+
+            self.pref_interval_by_bloc = pref_interval_by_bloc
+            self.bloc_voter_prop = bloc_voter_prop
+
+    @classmethod
+    def from_params(
```
jgibson517, line 59: I like this, its more user friendly then passing in a dictionary full of stuff. If a user sets the intervals when initializing are those overwritten when this method is called?

tests/test_models.py
```diff
@@ -1,147 +1,59 @@
...
-
-def test_gen_with_real_data():
-    model = DummyActual()
-    cands = {"R": ["A1", "B1", "C1"], "D": ["A2", "B2"]}
-    params = {}
-    with pytest.raises(TypeError):
-        model.generate_ballots(num_ballots=10, candidates=cands, hyperparams=params)
-
-
-def test_single_tie():
```
jgibson517, line 62: I don't think these tests should be deleted? they're checking input ties not ballot generation

jennjwang: will add back once we settle on ballot generation class

tests/test_ballot_generator.py
```diff
@@ -422,185 +445,6 @@ def group_elements_by_mapping(element_list, mapping):
     )
 
 
-# def test_Cambridge_distribution():
```
jgibson517, line 425: also looks like you overwrote some of the tweaks Zach and i made to make the cambridge test pass consistently, unless you two redid it after I left. looks like it was this in this commit: https://github.com/mggg/VoteKit/commit/798fef5c3296447c994ed04a20532c03f8a0469b

src/votekit/ballot_generator.py
```diff
@@ -1,70 +1,140 @@
...
+
+        if "pref_interval_by_bloc" not in data:
+            data["pref_interval_by_bloc"] = interval_by_bloc
+
+        if "bloc_voter_prop" not in data:
+            data["bloc_voter_prop"] = bloc_voter_prop
+
+        generator = cls(**data)
+
+        if isinstance(generator, AlternatingCrossover) or isinstance(
```
jamesturk, line 132: tip: second argument to isinstance can take a tuple `isinstance(generator, (AlternatingCrossover, CambridgeSampler))`

src/votekit/ballot_generator.py
```diff
@@ -1,70 +1,140 @@
...
-    # pref_interval_by_bloc: Optional[dict] = None  # race: {candidate : interval length}
-    # demo_breakdown: Optional[dict] = None  # race: percentage
-
     def __init__(
         self,
-        candidates: Union[list, dict],
+        candidates: list,
         ballot_length: Optional[int] = None,
-        hyperparams: Optional[dict] = {},
+        pref_interval_by_bloc=None,
```
jamesturk, line 21: consider making these arguments kw-only (https://peps.python.org/pep-3102/) by adding `, *,` before them

src/votekit/ballot_generator.py
```diff
@@ -183,17 +198,56 @@ def _preference_interval(
...
         perm_rankings = [list(value) for value in perm_set]
 
-        # IAC Process is equivalent to drawing from dirichlet dist with uniform parameters
-        draw_probabilities = np.random.dirichlet([self.alpha] * len(perm_rankings))
+        if self.alpha:
+            draw_probabilities = list(
+                np.random.dirichlet([self.alpha] * len(perm_rankings))
+            )
+        elif self.point:
+            draw_probabilities = [
```
jamesturk, line 238: this code gets surprisingly complex w/ nested list comprehensions, reduce, and conditionals in a lambda.  I'd add a comment to help walk readers through it

jamesturk (commented):
Deferring to Jack on the substance here since a lot is removed and not sure of the context.  Commented on a few tiny things.

src/votekit/ballot_generator.py
jennjwang: yes

PR #96, opened 2023-09-07
add new arguments

cdonnay:
Add a `rank_cols` argument to `load_csv` so that the user can specify which columns of the csv correspond to rankings. Helpful, for example, with the 2013 Minneapolis data file on MGGG/chicago which has extraneous columns.

Also removed `quota` parameter from `Election.get_threshold` method since `quota` should be an attribute of the Election.

Also, this is my first pull request; Moon has asked me to tag y'all going forward. Let me know if you don't wish to be tagged or if I made an error in my request. Thanks!
@jamesturk @drdeford @jgibson517 @ziglaser @jennjwang

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,4 @@ htmlcov/
 .DS_Store
 dist/
 .ipynb_checkpoints
-
+.idea
--- a/src/votekit/cvr_loaders.py
+++ b/src/votekit/cvr_loaders.py
@@ -11,17 +11,20 @@ from .ballot import Ballot
 
 def load_csv(
     fpath: str,
+    rank_cols: list[int],
     *,
     weight_col: Optional[int] = None,
     delimiter: Optional[str] = None,
     id_col: Optional[int] = None,
 ) -> PreferenceProfile:
     """
-    Given a file path, loads cvr with ranks as columns and voters as rows
+    Given a file path, loads cast vote records (cvr) with ranks as columns and voters as rows
     (empty cells are treated as None)
 
     Args:
         fpath: Path to cvr file
+        rank_cols: list of column indexes that contain rankings, indexing starts from 0,
+                    in order from top to bottom rank
         weight_col: The column position for ballot weights
             if parsing Scottish elections like cvrs
         delimiter: The character that breaks up rows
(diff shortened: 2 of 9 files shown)


jamesturk:
Hi @cdonnay,

A couple of notes:

1) You'll want to add .idea to .gitignore, and remove the .idea files, those are only relevant to you/PyCharm and shouldn't be part of this repo.
2) You're adding a required parameter, but it should probably be optional since it was already working without that.  This is also why the tests are failing.  (A new test should probably be added to accept this optional parameter.)

I'll let someone else comment on the self.quota issue as I haven't kept up with the API in that level of detail.

cdonnay:
Thanks, @jamesturk ! I added .idea to .gitignore, and gave the `rank_cols` a default argument. 

I think it's worth discussing whether or not you actually want the `rank_cols` parameter to have a default argument. My argument for not including a default is that we don't want to force preprocessing of the csv file onto the user, and we want the user to think intentionally about loading in their data. Happy to hear other ideas about why we might want a default.

jennjwang:
Hi @cdonnay thanks for helping with Votekit! I agree with James on making rank_cols an optional argument - I think the user should have an idea of what their data looks like before loading it, but it would be annoying for them to specify the rank_cols every time if their csv file is already clean without extraneous columns. A default would save them this trouble.

cdonnay:
Awesome, sounds good to me.

src/votekit/elections/election_types.py
```diff
@@ -61,20 +61,18 @@ def __init__(
         self.transfer = transfer
         self.seats = seats
         self.tiebreak = tiebreak
-        self.threshold = self.get_threshold(quota)
+        self.quota = quota.lower()
```
jgibson517, line 64: 👍 agree with this change

src/votekit/cvr_loaders.py
```diff
@@ -55,6 +59,12 @@ def load_csv(
     if id_col is not None and not df.iloc[:, id_col].is_unique:
         raise DataError(f"Duplicate value(s) in column at index {id_col}")
 
+    if rank_cols:
```
jgibson517, line 62: 👍

.gitignore
```diff
@@ -5,4 +5,4 @@ htmlcov/
 .DS_Store
 dist/
 .ipynb_checkpoints
-
+.idea
```
jgibson517, line 8: you may need to delete the .idea and inspectionProfiles files from the version of your repo on github, but then the .gitignore should handle them in the future!

cdonnay: Okay, I think I just deleted those files.

PR #97, opened 2023-09-07
plot_summary_stats x-axis

cdonnay:
Ensures that the candidates are always listed in the same order on the x-axis of `plot_summary_stats`.

### DIFF
--- a/src/votekit/plots/profile_plots.py
+++ b/src/votekit/plots/profile_plots.py
@@ -42,7 +42,10 @@ def plot_summary_stats(
 
     fig, ax = plt.subplots()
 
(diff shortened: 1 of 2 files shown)


src/votekit/plots/profile_plots.py
```diff
@@ -42,7 +42,10 @@ def plot_summary_stats(
 
     fig, ax = plt.subplots()
 
-    ax.bar(data.keys(), data.values(), color=colors, width=0.35)
+    candidates = profile.get_candidates()
```
jgibson517, line 45: 💯

jgibson517 (approved):

PR #99, opened 2023-09-13
Ballot graph

cdonnay:
Sorry for the length and number of changes here, but I was finding the `BallotGraph` object to not really be working as intended. High level summary of changes:

1. Change `complete` parameter to `allow_partial` in `__init__`. If `allow_partial` is True, BallotGraph constructs all possible ballot types. If False, it only generates full linear orderings.
2. Add `show_cast` parameter to `draw`. This only displays nodes of the graph that have the `cast` attribute equal to True.
3. Added/replaced necessary test functions.
4. Changed `node_data` to `node_weights` since that is the only kind of data it was storing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,4 @@ htmlcov/
 .DS_Store
 dist/
 .ipynb_checkpoints
-
+.idea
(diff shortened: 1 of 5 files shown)


cdonnay:
Fixed an error with the `show_cast` argument so that the `label_cands` method only generates labels for cast ballots.

cdonnay:
Closing this PR to start a new one with cleaner code.

PR #100, opened 2023-09-17
Ballot graph overhaul

cdonnay:
Sorry for the length and number of changes here, but I was finding the BallotGraph object to not really be working as intended. High level summary of changes:

1. Change `complete` parameter to `allow_partial` in `__init__`. If `allow_partial` is True, BallotGraph constructs all possible ballot types. If False, it only generates full linear orderings.
2. Add `fix_short` parameter to `__init__`. Gives the user the option to change ballots of length n-1 to n.
3. Add several parameters to `draw`. `to_display` lets the user specify a boolean function on nodes to be drawn. `neighborhoods` and `show_cast` are parameters that give common display functions.
4. Added/replaced necessary test functions.
5. Changed node_data to node_weights since that is the only kind of data it was storing.

### DIFF
--- a/src/votekit/graphs/ballot_graph.py
+++ b/src/votekit/graphs/ballot_graph.py
@@ -8,7 +8,7 @@ from functools import cache
 
 class BallotGraph(Graph):
     """
-    Class to build graphs for elections with possible incomplete ballots
+    Class to build ballot graphs.
 
     **Attributes**
 
@@ -16,8 +16,10 @@ class BallotGraph(Graph):
     :   data to create graph from, either PreferenceProfile object, number of
             candidates, or list of candidates
 
-    `complete`
-    :   if True, builds complete graph, else builds incomplete (boolean)
+    `allow_partial`
+    :   if True, builds graph using all possible ballots,
+        if False, only uses total linear ordered ballots
+        if building from a PreferenceProfile, defaults to True
 
     **Methods**
     """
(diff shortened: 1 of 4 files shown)


src/votekit/graphs/ballot_graph.py
```diff
@@ -185,9 +190,13 @@ def fix_short_ballot(self, ballot: list, candidates: list) -> list:
 
         return ballot + list(missing)
 
-    def label_cands(self, candidates):
+    def label_cands(self, candidates,
+                    to_display: Optional[Callable] = all_nodes):
```
jgibson517, line 194: Here's the issue with the type hinting as there's technically None being passed when this argument is not included. Changing it to `to_display: Optional[Union[Callable, None]]` I think will fix the linting issue. 

Separately, can you explain why this is a callable, what is the user supposed to input if they want to to only display a subset of the nodes? Is it a function that returns True/False for a given node?

cdonnay: Yeah, the callable should take in the graph and a node and return True if that node should be displayed. I just added an edit to make sure the graph was an argument to these callables.

cdonnay: Ah looks like the thing that fixed it was just using `Callable`. Now the only thing failing is `test_interval_from_params`, which is that non-deterministic test. Can you re-run it for me @jgibson517 ? Thanks!

cdonnay:
Updated the draw method to ensure labels do not go over the edge of the plot.

src/votekit/graphs/ballot_graph.py
jgibson517: 👍

PR #103, opened 2023-10-24
Ballot Generator Tutorials

cdonnay:
Add a tutorial notebook for the ballot generator models, with the exception of Alternating Crossover and Cambridge Sampler.  In creating the tutorial, I found that the PL and BT models could not handle the case when a candidate had 0 support from a bloc. I added code to handle this case. I also fixed how `PreferenceProfiles` are printed so that you can see ties.


cdonnay:
The PL and BT models do not always generate the exact number of ballots specified by the user due to rounding errors. To correct this, I implement Huntington Hill apportionment, which uses the `apportionment` package. I added that package to the `pyproject.yml` file

PR #104, opened 2023-10-31
AC and CS models

cdonnay:
Adding a notebook for the Alternating Crossover and Cambridge Sampler. Code changes include:

1. The AC model was not generating ballots correctly. As an example, in the old code, if a voter was a bloc voter, it just deterministically ranked the bloc candidates in front of the non-bloc candidates. It now uses the PL model to rank the bloc candidates within themselves.
2. I adapted the CS model to handle bloc names that differ from the historical Cambridge data.
3. I put in a print statement when printing PreferenceProfiles to let users know if it's been truncated.
4. In both the AC and CS model I implement Huntington-Hill to determine the number of ballots to generate.


src/votekit/ballot_generator.py
```diff
@@ -8,6 +8,7 @@
 import pickle
 import random
 from typing import Optional
+import apportionment.methods as apportion
```
jgibson517, line 11: Adding `# type: ignore` to end of this line should solve the mypy test failures.

src/votekit/ballot_generator.py
```diff
@@ -509,84 +515,110 @@ class AlternatingCrossover(BallotGenerator):
...
+        for bloc, pref_interval in self.pref_interval_by_bloc.items():
+            if 0 in pref_interval.values():
+                raise ValueError("In AC model, all candidates must have non-zero preference.")
 
     def generate_profile(self, number_of_ballots) -> PreferenceProfile:
 
         ballot_pool = []
 
+        # compute the number of bloc and crossover voters in each bloc using Huntington Hill
+        voter_types = [(b, type) for b in list(self.bloc_voter_prop.keys()) \
```
jgibson517, line 537: `self.bloc_voter_prop.keys()` returns a list, so there's no need to cast as a list here.

src/votekit/ballot_generator.py
```diff
@@ -509,84 +515,110 @@ class AlternatingCrossover(BallotGenerator):
...
+                
+                ballot = Ballot(ranking = ranking, weight = Fraction(1,1))
+                ballot_pool.append(ballot)
 
-                # Bloc ballots
-                for _ in range(num_ballots - num_crossover_ballots):
-                    ballot = bloc_cands + opposing_cands
-                    ballot_pool.append(ballot)
+            for _ in range(num_bloc_ballots):
+                bloc_cands = list(
```
jgibson517, line 599: Let's create a helper function/method to create `bloc_cands` and `opposing_cands` instead of copy/pasting that code!

src/votekit/ballot_generator.py
```diff
@@ -509,84 +515,110 @@ class AlternatingCrossover(BallotGenerator):
...
+            # convert to probability distribution
+            pref_for_opposing = [
+                p / sum(pref_for_opposing) for p in pref_for_opposing
+            ]
+
+            pref_for_bloc = [pref_interval_dict[cand] for cand in bloc_cands]
+            # convert to probability distribution
+            pref_for_bloc = [p / sum(pref_for_bloc) for p in pref_for_bloc]
+
+            for _ in range(num_cross_ballots):
```
jgibson517, line 572: Why are there two for loops (one for `range(num_cross_ballots)` and another  for `range(num_cross_ballots)`), it looks like the `ranking` is all that is changing between them, but correct me if wrong!

Could we do `for i in range (num_cross_ballots + num_bloc_ballots)` and then use a conditional based on the number of ballots generated to switch whether the ranking is a crossover ballot or an opposing?

src/votekit/ballot_generator.py
```diff
@@ -509,84 +515,110 @@ class AlternatingCrossover(BallotGenerator):
...
-                            item
-                            for pair in zip(opposing_cands, bloc_cands)
-                            for item in pair
-                            if item is not None
-                        ]
-
-                    # check that ballot_length is shorter than total number of cands
-                    ballot_pool.append(ballot)
+                # alternate the bloc and opposing bloc candidates to create crossover ballots
+                ranking = [{cand} for pair in zip(opposing_cands, bloc_cands)\
```
jgibson517, line 592: Per my above comment, using an `if` statement here to switch how we create the ranking based on what iteration (or number of ballots generated) we are on in the for loop.

src/votekit/ballot_generator.py
```diff
@@ -644,16 +683,37 @@ class CambridgeSampler(BallotGenerator):
...
+        self.cohesion_parameters = cohesion_parameters
+
+        # changing names to match historical data
+        majority_bloc = [bloc for bloc, prop in self.bloc_voter_prop.items() if prop>=.5][0]
+        minority_bloc = [bloc for bloc in self.bloc_voter_prop.keys() \
+                                              if bloc != majority_bloc][0]
+        
+        cambridge_names = {majority_bloc: majority_name, minority_bloc: minority_name}
+
+        self.slate_to_candidates = {cambridge_names[b]: self.slate_to_candidates[b] \
```
jgibson517, line 706: This is more a style thing but you don't need the line break `\` since everything is inside the comprehension\brackets `{}`

src/votekit/ballot_generator.py
```diff
@@ -644,16 +683,37 @@ class CambridgeSampler(BallotGenerator):
     def __init__(
         self,
         slate_to_candidates=None,
-        bloc_crossover_rate=None,
+        cohesion_parameters=None,
         path: Optional[Path] = None,
+        majority_name: Optional[str] = "W",
```
jgibson517, line 688: Smart!

notebooks/AC_and_CS.ipynb
```diff
@@ -0,0 +1,200 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import votekit.ballot_generator as bg"
```
jgibson517, line 9: Let's move this line `import votekit.ballot_generator as bg` down to first big code chunk just for cohesiveness

jgibson517 (changes requested):
Hi Chris! Apologies for the delayed review on this PR. Thanks again for the help with documentation/demo noteboooks!

A couple of small comments on the changes to `ballot_generator.py` but otherwise looks good!

src/votekit/ballot_generator.py
cdonnay: I think with the addition of the if statement, this code isn't repeated and so we can just leave it as two calls to the `np.random` module.

PR #106, opened 2023-11-08
Cumulative voting

cdonnay:
This PR adds functionality for points based ballots. I felt like this deserved a new class, as it really is distinct from full rankings. 

- add `PointBallot` class
- adapt `PreferenceProfile` to handle new ballot class
- add `Cumulative` ballot generator class to create ballots for cumulative style elections.
- add a `HighestScore` election that takes a `PreferenceProfile` of `PointBallot`s and chooses the $m$ candidates with highest scores.

### DIFF
--- a/src/votekit/ballot_generator.py
+++ b/src/votekit/ballot_generator.py
@@ -9,7 +9,7 @@ import pickle
 import random
 from typing import Optional
 
-from .ballot import Ballot, CumulativeBallot
+from .ballot import Ballot, PointBallot
 from .pref_profile import PreferenceProfile
 import apportionment.methods as apportion
 
@@ -776,15 +776,20 @@ class Cumulative(BallotGenerator):
     `bloc_voter_prop`
     :   dictionary mapping of slate to voter proportions (ex. {race: voter proportion})
 
+    `num_votes`
+    : the number of votes an individual voter has, usually the same as the number of seats up for 
+    election.
+
     **Methods**
 
     See `BallotGenerator` base class
     """
 
-    def __init__(self, **data):
+    def __init__(self, num_votes: int = 1, **data):
 
         # Call the parent class's __init__ method to handle common parameters
         super().__init__(**data)
+        self.num_votes = num_votes
 
     def generate_profile(self, number_of_ballots) -> PreferenceProfile:
         ballot_pool = []
@@ -811,13 +816,13 @@ class Cumulative(BallotGenerator):
                 ballot = list(
                     np.random.choice(
                         non_zero_cands,
-                        len(non_zero_cands),
+                        self.num_votes,
                         p=cand_support_vec,
                         replace=True,
                     )
                 )
 
-                ballot = CumulativeBallot(multi_votes=ballot)
+                ballot = PointBallot(points=ballot)
                 ballot_pool.append(ballot)
             
 
--- a/src/votekit/ballot_generator.py
+++ b/src/votekit/ballot_generator.py
@@ -11,7 +11,7 @@ from typing import Optional
 
 from .ballot import Ballot, PointBallot
 from .pref_profile import PreferenceProfile
-import apportionment.methods as apportion
+import apportionment.methods as apportion # type: ignore
 
 class BallotGenerator:
     """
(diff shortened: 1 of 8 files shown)


cdonnay:
Looks like creating this new `PointBallot` class has thrown a lot of type errors in `mypy`. Would be happy to chat about the best way to address that!

jgibson517:
Hi @cdonnay, I view the `Ballot` object as more of data structure then an object that dictates what type of ranking system is used. Instead I might lean towards adding an optional `points` attribute to the existing class.

A couple thoughts here: When we first designed the class, I think we didn't think users would really need to touch it much, but I don't think that's really the case anymore. So I think if we give a little more flexibility in how we initialize a Ballot, we won't need a second class.  

I think an optional `points` method in the form of dictionary or vector can live in Ballot class. I also like in your `PointBallot` you default the weights to 1 if a user does not specify, we should adopt that into the `Ballot` object. Since the rest of ballot functionality is not exposed to the user (i.e users don't adjust ballot attributes when running elections), I think its more streamlined to have a single Python object do the work.  Happy to chat about this more too, if helpful!!

src/votekit/ballot_generator.py
```diff
@@ -9,9 +9,9 @@
 import random
 from typing import Optional
 
-from .ballot import Ballot
+from .ballot import Ballot, PointBallot
 from .pref_profile import PreferenceProfile
-
+import apportionment.methods as apportion # type: ignore
```
jamesturk, line 14: to ensure this works as a package, this should be a relative import like the ones above (start with one or more `.` to define the relative path between them

src/votekit/elections/election_types.py
```diff
@@ -900,3 +901,57 @@ def __init__(
...
+    def run_step(self):
+        # dictionary whose keys are candidates, values are vote totals
+        vote_tallies = self.state.profile.to_dict()
+
+        ranking = scores_into_set_list(score_dict = vote_tallies)
+        untied_ranking = tie_broken_ranking(ranking = ranking, profile = self.state.profile,
+                                     tiebreak=self.tiebreak)
+
+        if ranking != untied_ranking:
+            print("A tie was broken.")
```
jamesturk, line 941: probably want to avoid `print` here, maybe `logging` or adding some other way of checking this after the fact if desired

src/votekit/models.py
```diff
@@ -13,7 +14,7 @@ class Election(ABC):
     """
 
     def __init__(self, profile: PreferenceProfile, ballot_ties: bool = True):
-        if ballot_ties:
+        if ballot_ties and isinstance(profile.ballots[0], Ballot):
```
jamesturk, line 17: I think a comment here to clarify this logic would be helpful

src/votekit/pref_profile.py
```diff
@@ -63,27 +68,46 @@ def num_ballots(self) -> Fraction:
...
         """
         num_ballots = self.num_ballots()
         di: dict = {}
-        for ballot in self.ballots:
-            rank_tuple = tuple(next(iter(item)) for item in ballot.ranking)
-            if standardize:
-                weight = ballot.weight / num_ballots
-            else:
+
+        if isinstance(self.ballots[0], Ballot):
```
jamesturk, line 86: a lot of shared code between these two, moot if they merge, but I'd refactor this to avoid duplication if not

src/votekit/pref_profile.py
```diff
@@ -95,29 +119,52 @@ def to_csv(self, fpath):
         Args:
             fpath (str): path to the saved csv
         """
-        with open(fpath, "w", newline="") as csvfile:
-            fieldnames = ["weight", "ranking"]
-            writer = csv.DictWriter(csvfile, fieldnames=fieldnames)
-            writer.writeheader()
-            for ballot in self.ballots:
-                writer.writerow({"weight": ballot.weight, "ranking": ballot.ranking})
+        if isinstance(self.ballots[0], Ballot):
```
jamesturk, line 122: same as above comment, lots of duplication here when the two types can probably share most of the logic

jamesturk (commented):
Left some inline comments, overall I'd agree with @jgibson517 that it'd make sense to combine these into one type to avoid the duplication.  The places where they differ could be smoothed out by adding additional helper methods onto `Ballot` itself, or within `Cumulative`

src/votekit/ballot_generator.py
jamesturk: nevermind, thought this was an internal module

jamesturk:
re: mypy errors
If this does go forward as two separate classes, `PointBallot` should probably inherit from `Ballot` which will help with the type issues but also break the current `isinstance` checks.   It'd be preferred here to have the two share an explicit interface, which is part of what `mypy` is so upset about.  Once that's true they can be used *almost* interchangably, relying on checking for `PointBallot` only when its customizations are needed.

cdonnay:
Alright, let me take another go at this using just a `Ballot` class, and implementing some kind of point system within that.

PR #107, opened 2023-11-10
Adding to_dict() to election state class

jgibson517:
- added to_dict(), to_json() functions for election_state
- updated election_state methods to better express return values
- caches for election_types


jgibson517:
Currently saving output ties as sets, so that the dictionary is not serializable with saving to JSON :/

jgibson517:
@cdonnay I changed some of the methods names in outcome class what do you think? // want to make sure any tutorials reflect the updated names!

jamesturk:
It's probably a good idea to start maintaining a changelog.md in the docs so that people (including the core team, because you will forget) can figure out what changed in a given version and how to update their code.

https://keepachangelog.com/en/1.1.0/ has good guidelines, and pretty much any package you're using (pandas, python itself) have solid examples.  You typically want to at least list breaking changes clearly, while documenting bugfixes and additions well is useful too.

cdonnay (commented):
Love the new names, looks good to me.

PR #109, opened 2023-11-14
Update Documentation

cdonnay:
Tries to make the documentation clearer and more consistent across the API page.

1. Is there any way I can preview what the API page will look like?
2. It feels like we need two kinds of documentation; Python documentation for VoteKit, and social choice theory documentation to explain our models. For the latter kind, should I just make some markdown notebooks?


jamesturk:
If you run `poetry run mkdocs serve` it should give you a live-updating preview of all changes as you make them.

re: #2, some social choice theory documentation seems like it'll be important and a great idea. Others can weigh in but I personally think it'd be nice to have it all in one place so it can crosslink easily/etc.

(e.g. add some markdown pages to docs/ and update mkdocs.yml accordingly, those docs can then easily reference the API documentation and vice versa)

cdonnay:
I fixed some lingering capitalization errors and did some formatting on the API page:

1. Indented the table of contents so that the headers are visible.
2. Removed `round_num` and `unset` from the documentation as users do not  need to see these helper functions.

jgibson517:
Also agree that it would be smart to have both the theory and code documentation in the same place! @cdonnay let's put our heads together to divvy up that work

PR #113, opened 2023-11-19
Stv bugs

jgibson517:
fixing some lingering bugs in the STV class:
- add scores attribute to ElectionState to keep track of scores/votes from previous rounds
- multiple round winners are now correctly ranked based on the last rounds scores

Ballot objects are now **immutable**:
- in practice, ballots had been treated as immutable but this is now enforced through a dataclass
- `get_profile()` method in PreferenceProfile now returns a sliced copy of ballots, so that election classes no longer modify the PreferenceProfile in place


cdonnay (commented):
These look good to me! Looks like we have some conflicts in the `Ballot` class: I had changed the name of `voters` to `voter_set` to be a bit more descriptive, and added some `__init__` code to handle the user inputting a non-Fraction object.

jgibson517:
@cdonnay I added a post_init function that does the conversion from non-Fraction types to Fractions, and plays a little nicer with the pydantic datatclass. Also added a changelog.md, I migrated the github release notes to this file to get us started, but we can get in a better practice of documenting changes going forward!

PR #114, opened 2023-11-25
ballot class updates

cdonnay:
Updates to ballot class for tutorial notebooks.


cdonnay:
Edits approved and viewed by Moon.

PR #117, opened 2023-11-29
enable installation on Python 3.12

jamesturk:
- bump supported and tested version to 3.12
- add 3.12 to changelog


cdonnay:
supported!

PR #118, opened 2023-12-01
Enable points based voting

cdonnay:
Here are some updates that allow for points based voting methods, as well as some general features.

1. Added a `by_bloc` parameter to `generate_profile`. If True, this returns a dictionary of PreferenceProfiles by bloc. This is very helpful for analyzing the behavior of a single bloc of voters. Defaults to False for backwards compatibility.
2. Created a `Cumulative` ballot generator class. The `Cumulative` class works like PL, but samples with replacement instead of without. The ranking order does not matter here, simply that candidates are listed on the ballot with multiplicity.
3. Created a `HighestScore` election class. This takes in a profile and a score vector, and returns the $m$ candidates with highest scores. There is a lot of flexibility in the score vector, so this class can run things like Borda, cumulative, etc.
4. Created a `Cumulative` election class which is just a subclass of `HighestScore` with the score vector set to all 1s. Thus anyone appearing on the ballot gets one point for each time they appear.
5. Wrote an `__add__` method for `PreferenceProfile` that combines the ballot lists of two profiles.
6. Created utility functions to compute the winners of a profile given a score vector, as well as to validate a score vector (non-negative and non-increasing).

I did not make `Borda` a subclass of `HighestScore` since it seems like there are some finer details about how short ballots/overvotes are handled in the existing `Borda` class.


cdonnay:
1. Added a `Limited` ballot generator, which operates just like PL but for when the ranking is not necessarily full.
2. Made `PlackettLuce` a subclass of `Limited`.

src/votekit/ballot_generator.py
```diff
@@ -977,3 +910,247 @@ def generate_profile(self, number_of_ballots: int) -> PreferenceProfile:
...
+        
+        if by_bloc:
+            return pp_by_bloc
+
+        # else combine the profiles
+        else:
+            pp = PreferenceProfile(ballots=[])
+            for profile in pp_by_bloc.values():
+                pp+= profile
+            return(pp)
```
drdeford, line 1124: This is an elegant approach but if people want the by_bloc version are they also likely to want the regular profile? And if so would it be better to return both?

src/votekit/elections/election_types.py
```diff
@@ -951,3 +951,117 @@ def __init__(
...
+    `score_vector`
+    : list of floats where ith entry denotes the number of points given to candidates
+        ranked in position i.
+
+    `tiebreak` (optional)
+    : tiebreak method, defaults to random, supports 'none', 'random',
+
+    `ballot_ties` (optional)
+    : resolves ties in ballots. Should only be set to True if you want ballots
+        to have full linear rankings.
```
drdeford, line 978: Would it be better for this to accept a method to make it more flexible for users?

src/votekit/pref_profile.py
```diff
@@ -303,3 +300,15 @@ def format_as_float(percent_str):
...
+
+    def __add__(self, other):
+        """
+        Add two PreferenceProfiles by combining their ballot lists.
+        """
+        if isinstance(other, PreferenceProfile):
+            ballots = self.ballots+other.ballots
+            pp = PreferenceProfile(ballots=ballots)
+            pp.condense_ballots()
+            return pp
```
drdeford, line 312: Not sure what the current protocol is but this seems like something that could use some tests - easy to imagine places where the condensing leads to unintuitive results.

src/votekit/utils.py
```diff
@@ -328,6 +330,51 @@ def scores_into_set_list(
...
+
+    for ballot in profile.ballots:
+        for i, s in enumerate(ballot.ranking):
+            # for each candidate in position i, give them points as determined by score_vector
+            for c in s:
+                try:
+                    candidates_to_scores[c] += score_vector[i]*ballot.weight
+                except IndexError:
+                    raise IndexError(f"Tried to access index {i} of score vector," 
+                                     f"but vector only length {len(score_vector)}.")
```
drdeford, line 359: Would it be better to assume unassigned/inaccessible ranks are 0 and raise a warning instead?

cdonnay: This seems reasonable to me.

cdonnay:
1. Added tests for add method of PreferenceProfile.
2. Raised a warning in computing scores from score vectors.
3. Renamed Limited class to shortPlackettLuce.

src/votekit/elections/election_types.py
cdonnay: If a user wanted to apply their own ballot tie resolving method, they could pass it to the existing `clean_profile` function before passing the profile into the election class, so I think we can leave the `ballot_ties` as is!

I do think it would be nice for users to be able to pass in their own tiebreak method, so I'll make that change.

src/votekit/ballot_generator.py
cdonnay: I think you're right!

PR #119, opened 2023-12-07
Some post tutorial updates

cdonnay:
1. Change the way the `condense_ballots()` method works in profiles. Rather than altering the original profile, it returns a new profile. This gives users the option to preserve the original profile.
2. Add an IRV election class, which is just a wrapper for STV with 1 seat.
3. Add default option to Borda election class, so users do not have to input a score vector if they want to use the traditional Borda vector.
4. Alter STV class so that the remaining candidates are always listed in order of current first place votes.
5. Add print statement to `BallotGraph` so that when you draw the graph without labels, it prints a dictionary of candidate labels for you.
6. Add several methods to `PairwiseComparisonGraph`. Added two boolean methods that return True if there is a condorcet winner or if there is a condorcet cycle. Added two get methods that return the winner or the cycles. Cached the results of `dominating_tiers` and `get_condorcet_cycles`.
7. Fixed an error in the `PreferenceProfile` tail method.
8. Added optional `to_float` method to `first_place_votes` if users want to see them as floats instead of Fractions.

I will add these to the change log only once the PR has been merged.

### DIFF
--- a/src/votekit/election_state.py
+++ b/src/votekit/election_state.py
@@ -91,8 +91,9 @@ class ElectionState(BaseModel):
         """
         if self.curr_round == round:
             return {
-                "Elected": [c for s in self.elected for c in s],
-                "Eliminated": [c for s in self.eliminated_cands for c in s],
+                "Elected": [s for s in self.elected ],
+                "Eliminated": [s for s in self.eliminated_cands],
+                "Remaining": [s for s in self.remaining]
             }
         elif self.previous:
             return self.previous.round_outcome(round)
@@ -150,10 +151,17 @@ class ElectionState(BaseModel):
 
         for round in range(1, self.curr_round + 1):
             results = self.round_outcome(round)
-            for status, candidates in results.items():
-                for cand in candidates:
-                    status_df.loc[status_df["Candidate"] == cand, "Status"] = status
-                    status_df.loc[status_df["Candidate"] == cand, "Round"] = round
+            for status, ranking in results.items():
+                for s in ranking:
+                    for cand in s:
+                        tied_str = ""
+                        # if tie
+                        if len(s) > 1:
+                            remaining_cands = ", ".join(list(s.difference(cand)))
+                            tied_str = f" (tie with {remaining_cands})"
+                        
+                        status_df.loc[status_df["Candidate"] == cand, "Status"] = status + tied_str
+                        status_df.loc[status_df["Candidate"] == cand, "Round"] = round
 
         return status_df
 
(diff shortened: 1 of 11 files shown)


cdonnay:
Further edits:

1. Fixed error in STV class where if two candidates were elected simultaneously, the fractional transfer would occur twice.
2. Added a tie string to the printing of `ElectionState` so users can see if candidates were tied when elected.

src/votekit/election_state.py
```diff
@@ -91,8 +91,9 @@ def round_outcome(self, round: int) -> dict:
         """
         if self.curr_round == round:
             return {
-                "Elected": [c for s in self.elected for c in s],
-                "Eliminated": [c for s in self.eliminated_cands for c in s],
+                "Elected": [s for s in self.elected ],
```
jamesturk, line 94: nit: I'd probably use list(self.elected) here for clarity and a tiny speed improvement

src/votekit/pref_profile.py
```diff
@@ -261,28 +265,24 @@ def condense_ballots(self):
...
         if not isinstance(other, PreferenceProfile):
             return False
-        self.condense_ballots()
-        other.condense_ballots()
-        for b in self.ballots:
-            if b not in other.ballots:
-                return False
-        for b in self.ballots:
-            if b not in other.ballots:
+        pp_1 = self.condense_ballots()
```
jamesturk, line 282: does this change neglect case where pp_2 has a ballot not in pp_1?

jamesturk (commented):
Mostly looks good, flagged one small thing and one potential error if I'm not misunderstanding context.

I'd also recommend continuing to run black/ruff on this, as there are quite a few deviations from PEP8 style that they'll pick up.

cdonnay:
@jamesturk  thanks for the edits!

I'm currently running 
- poetry run mypy src
- poetry run ruff src tests
- poetry run pytest
Is there another one I should do?

src/votekit/graphs/pairwise_comparison_graph.py
```diff
@@ -216,3 +232,32 @@ def dominating_tiers(self) -> list[set]:
...
+        Returns a list of condorcet cycles in the graph, which we define as any cycle of length
+            greater than 2.
+
+        Returns:
+            List of condorcet cycles
+        """
+
+        G = self.pairwise_graph
+
+        return(nx.recursive_simple_cycles(G))
```
drdeford, line 263: Do we want to return these sorted by e.g. length?

cdonnay: I think that's a great idea.

PR #120, opened 2024-01-19
add SlatePreference model

cdonnay:
This PR adds a new model of ballot generator, the SlatePreference model, as well as two tests for the test suite.

1. The SlatePreference model uses the cohesion parameter of a voter to determine their ballot type; a coin weighted by the cohesion parameter is flipped repeatedly, determining the order of the slates on the ballot. The candidate names are then filled in in a PL style.
2. Added completion and distribution tests for the SP model.

### DIFF
--- a/src/votekit/ballot_generator.py
+++ b/src/votekit/ballot_generator.py
@@ -453,8 +453,8 @@ class BradleyTerry(BallotGenerator):
             ranking_to_prob[ranking] = prob
         return ranking_to_prob
 
-    def generate_profile(self, number_of_ballots, by_bloc: bool = False) -> Union[PreferenceProfile,
-                                                                                  Tuple]:
+    def generate_profile(self, number_of_ballots: int, by_bloc: bool = False) \
+                                                                -> Union[PreferenceProfile, Tuple]:
         
 
         # the number of ballots per bloc is determined by Huntington-Hill apportionment
@@ -585,8 +585,8 @@ class AlternatingCrossover(BallotGenerator):
                     "In AC model, all candidates must have non-zero preference."
                 )
 
-    def generate_profile(self, number_of_ballots, by_bloc: bool = False) -> Union[PreferenceProfile,
-                                                                                  Tuple]:
+    def generate_profile(self, number_of_ballots: int, by_bloc: bool = False) \
+                                                                -> Union[PreferenceProfile, Tuple]:
         
 
         # compute the number of bloc and crossover voters in each bloc using Huntington Hill
@@ -696,8 +696,8 @@ class OneDimSpatial(BallotGenerator):
     See `BallotGenerator` base class.
     """
 
-    def generate_profile(self, number_of_ballots, by_bloc: bool = False) -> Union[PreferenceProfile,
-                                                                                  dict]:
+    def generate_profile(self, number_of_ballots: int, by_bloc: bool = False) \
+                                                                -> Union[PreferenceProfile, Tuple]:
         candidate_position_dict = {c: np.random.normal(0, 1) for c in self.candidates}
         voter_positions = np.random.normal(0, 1, number_of_ballots)
 
@@ -820,8 +820,8 @@ class CambridgeSampler(BallotGenerator):
             for b in self.cohesion_parameters.keys()
         }
 
-    def generate_profile(self, number_of_ballots, by_bloc: bool = False) -> Union[PreferenceProfile,
-                                                                                  Tuple]:
+    def generate_profile(self, number_of_ballots: int, by_bloc: bool = False) \
+                                                                -> Union[PreferenceProfile, Tuple]:
         with open(self.path, "rb") as pickle_file:
             ballot_frequencies = pickle.load(pickle_file)
 
@@ -988,8 +988,8 @@ class Cumulative(BallotGenerator):
         self.num_votes = num_votes
 
 
-    def generate_profile(self, number_of_ballots, by_bloc: bool = False) -> Union[PreferenceProfile,
-                                                                                  Tuple]:
+    def generate_profile(self, number_of_ballots: int, by_bloc: bool = False) \
+                                                                -> Union[PreferenceProfile, Tuple]:
         """
         Args:
         `number_of_ballots`: The number of ballots to generate.
@@ -1083,8 +1083,8 @@ class shortPlackettLuce(BallotGenerator):
         super().__init__(**data)
         self.k = k
 
-    def generate_profile(self, number_of_ballots, by_bloc: bool = False) -> Union[PreferenceProfile,
-                                                                                  Tuple]:
+    def generate_profile(self, number_of_ballots: int, by_bloc: bool = False) \
+                                                                -> Union[PreferenceProfile, Tuple]:
         """
         Args:
         `number_of_ballots`: The number of ballots to generate.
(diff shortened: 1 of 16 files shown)


src/votekit/ballot_generator.py
```diff
@@ -820,8 +820,8 @@ def _rename_blocs(self):
             for b in self.cohesion_parameters.keys()
         }
 
-    def generate_profile(self, number_of_ballots, by_bloc: bool = False) -> Union[PreferenceProfile,
-                                                                                  Tuple]:
```
jamesturk, line 824: just a question, is there a reason for this change? since the project is still using `black` for formatting, these will be changed back once it is run.

src/votekit/ballot_generator.py
```diff
@@ -1201,4 +1201,144 @@ def __init__(self, **data):
...
+
+    `cohesion_parameters`
+    :   dictionary mapping of slate to cohesion level (ex. {bloc: .7}).
+
+    **Methods**
+
+    See `BallotGenerator` base class
+    """
+
+    def __init__(self, slate_to_candidates: dict = {}, cohesion_parameters: dict = {}, **data):
```
jamesturk, line 1238: these default parameters should not be {}, see https://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python for a good explanation as to why.

This will lead to strange hard-to-diagnose bugs.  Instead default to `None`, and have a conditional inside the constructor that sets to `{}` if `None` was passed.

src/votekit/ballot_generator.py
```diff
@@ -1201,4 +1201,144 @@ def __init__(self, **data):
...
+        )
+
+        pref_profile_by_bloc = {}
+
+        for bloc in blocs: 
+            ballot_pool = []
+            # number of voters in this bloc
+            num_ballots = ballots_per_block[bloc]
+
+            # TODO the code below assumes only two blocs
```
jamesturk, line 1260: maybe add an exception if more than 2 bloc are passed in right now so code doesn't silently give incorrect results for >2 bloc

src/votekit/ballot_generator.py
```diff
@@ -1201,4 +1201,144 @@ def __init__(self, **data):
...
+
+            opp_bloc = set(blocs).difference(bloc).pop()
+            
+            for _ in range(num_ballots):
+                # sample ballot type
+                ballot_type = []
+                candidate_count = {b:0 for b in blocs}
+                for _ in range(len(non_zero_cands)):
+                    # choose a bloc as the next slot in the ballot
+                    cand_bloc = list(np.random.choice([bloc, opp_bloc], size =1,
```
jamesturk, line 1273: recommend to run autoformatter to fix style issues

src/votekit/ballot_generator.py
```diff
@@ -1201,4 +1201,144 @@ def __init__(self, **data):
...
+                    candidate_count[cand_bloc] += 1
+
+                    # if there are more candidates to fill
+                    if len(ballot_type) < len(non_zero_cands):
+                        for b in blocs:
+                            # and you are out of non-zero candidates for a slate
+                            if candidate_count[b] == len([c for c in self.slate_to_candidates[b] 
+                                                        if self.pref_interval_by_bloc[bloc][c]>0]):
+                                 # TODO this assumes only two blocs
+                                 o_bloc = set(blocs).difference(b).pop()
```
jamesturk, line 1285: same note as above regarding failing loudly if there are multiple blocs

src/votekit/ballot_generator.py
```diff
@@ -1201,4 +1201,144 @@ def __init__(self, **data):
...
+                    # sample
+                    cand_ordering = np.random.choice(cands, len(cands), 
+                                                     replace=False, p = distribution)
+                    cand_ordering_by_bloc[b] = list(cand_ordering)
+                
+                ranking = []
+                for b in ballot_type:
+                    # append the current first candidate, then remove them from the ordering
+                    ranking.append({cand_ordering_by_bloc[b][0]})
+                    cand_ordering_by_bloc[b].pop(0)
```
jamesturk, line 1322: small thing: pop(0) returns 0th element, you can combine these lines if you want

src/votekit/ballot_generator.py
```diff
@@ -1201,4 +1201,144 @@ def __init__(self, **data):
...
+                
+                zero_cands = [c for c,s in pref_interval.items() if s == 0]
```
jamesturk, line 1324: swap [] for {} and this becomes a set comprehension, so no need to cast to set on line below

src/votekit/ballot_generator.py
```diff
@@ -1201,4 +1201,144 @@ def __init__(self, **data):
...
+            ballot_pool = []
+            # number of voters in this bloc
+            num_ballots = ballots_per_block[bloc]
+
+            # TODO the code below assumes only two blocs
+            ballot_type_distribution = [self.cohesion_parameters[bloc], 
+                                        1-self.cohesion_parameters[bloc]]
+            non_zero_cands = [c for c,v in self.pref_interval_by_bloc[bloc].items() if v>0]
+
+            opp_bloc = set(blocs).difference(bloc).pop()
```
jamesturk, line 1265: ?

src/votekit/ballot_generator.py
```diff
@@ -1201,4 +1201,144 @@ def __init__(self, **data):
...
+        # Call the parent class's __init__ method to handle common parameters
+        super().__init__(**data)
+        self.slate_to_candidates = slate_to_candidates
+        self.cohesion_parameters = cohesion_parameters
+    
+    def generate_profile(self, number_of_ballots: int, by_bloc: bool = False) \
+                                                            -> Union[PreferenceProfile, Tuple]:
+        # the number of ballots per bloc is determined by Huntington-Hill apportionment
+        blocs = list(self.bloc_voter_prop.keys())
+        bloc_props = list(self.bloc_voter_prop.values())
```
jamesturk, line 1248: note: since these methods are called a lot, I'm going to point out small speed issues that might accumulate.  feel free to ignore since most are quite small, but they could add up in practice.

these two casts to list probably aren't needed, and if they can be removed will save memory/time

tests/test_ballot_generator.py
```diff
@@ -48,7 +49,20 @@ def test_PL_completion():
...
+        slate_to_candidates={"W": ["W1", "W2"], "C": ["C1", "C2"]},
+        pref_interval_by_bloc={
+            "W": {"W1": 0.4, "W2": 0.3, "C1": 0.2, "C2": 0.1},
+            "C": {"W1": 0.2, "W2": 0.2, "C1": 0.3, "C2": 0.3},
+        },
+        bloc_voter_prop={"W": 0.7, "C": 0.3},
+        cohesion_parameters={"W": 0.7, "C": 0.9},
+    )
+    profile = pl.generate_profile(number_of_ballots=100)
+    assert type(profile) is PreferenceProfile
```
jamesturk, line 65: should test verify that the PreferenceProfile is the expected one here?

jamesturk (commented):
👋 Left some comments, most fairly trivial but a few, including a default parameter issue that could cause some trouble left unchecked.

src/votekit/ballot_generator.py
cdonnay: Actually, I'm not sure what I was thinking on this one. I'm sure I was trying to avoid some error with the 100 character limit on lines, but yeah the comma situation seems cleaner than a line break.

src/votekit/ballot_generator.py
cdonnay: Turns out these shouldn't be optional anyway, but noted for the future!

src/votekit/ballot_generator.py
cdonnay: Cleaned up!

tests/test_ballot_generator.py
cdonnay: Good call

src/votekit/ballot_generator.py
cdonnay: It's not super easy to avoid casting them as a list, since the apportionment package needs `bloc_props` as a list. Moreover, these number of blocs is almost always going to be small so the cloning here shouldn't be too expensive.

I think the bigger fix here is to create an attribute `self.blocs` in the init method of all of these ballot generators since it get's used so often, and then at least it's only once per object, not once per call to generator.

PR #125, opened 2024-04-04
Fix SPL zero support candidate, edit Dirichlet notebook

cdonnay:
- Dirichlet parameter tutorial notebook now uses slate-PlackettLuce.

- Fixed bug by which slate-PlackettLuce could not generate ballots when some candidate had 0 support. Added two tests to catch.


peterrrock2 (approved):
Looks good!

PR #127, opened 2024-06-06
Update CS blocs

cdonnay:
Allow users to choose which bloc corresponds to W in historical Cambridge data. Default behavior is the same; whichever bloc is the majority in `bloc_voter_prop` is assigned to W.


peterrrock2 (approved):
Looks good!

PR #128, opened 2024-06-30
Update load_scottish

cdonnay:
Since the BLT files in the scot-elex repo were edited to be true CSV files, `load_scottish` has been updated to account for that.


src/votekit/cvr_loaders.py
peterrrock2, line 1: The `load_scottish` function should really make use of the csv module to help avoid parsing errors. Here is some example code that you can use:


```
import csv

# Convert the ballot rows to ints while leaving the candidates as strings
def convert_row(row):
    return [int(item) if item.isdigit() else item for item in row]

with open("test_election.csv", 'r') as f:
    reader = csv.reader(f)
    data = []
    for row in reader:
        # This just removes any empty strings that are hanging out since
        # we don't need to preserve columns
        filtered_row = list(filter(lambda x: x != '', row))
        data.append(convert_row(filtered_row))
```

peterrrock2 (changes requested):

peterrrock2 (approved):
This looks great!

PR #130, opened 2024-07-03
random elections and spatial generation

kevin-q2:
Initial work on random election mechanisms and spatial generation of preference profiles. Includes new election mechanisms RandomDictator and BoostedRandomDictator as well as Uniform/Clustered spatial generation classes.


cdonnay (changes requested):
Hi Kevin, this is great. My big comment is that I think we can refactor some stuff and add a little more flexibility to these things. Let's Zoom to strategize.

 @peterrrock2 , can you take a look at where I tagged you? I'll save you from a bigger review until after we refactor.

src/votekit/ballot_generator.py
```diff
@@ -1930,3 +1930,140 @@ def generate_profile(
...
+
+
+class UniformSpatial(BallotGenerator):
+    """
+    Uniform spatial model for ballot generation. Assumes the candidates are uniformly distributed in
+    d dimensional space. Voters are then distributed equivalently, and vote with
+    ballots based on Euclidean distance to the candidates.
+
+    Args:
+        candidates (list): List of candidate strings.
```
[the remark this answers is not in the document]
cdonnay: let's pass all distributions and distances here, instead of to generate_profile.

src/votekit/ballot_generator.py
```diff
@@ -1930,3 +1930,140 @@ def generate_profile(
...
+            Union[PreferenceProfile, Tuple]
+        """
+
+        np.random.seed(seed)
+        candidate_position_dict = {
+            c: np.random.uniform(lower, upper, size=dim) for c in self.candidates
+        }
+        voter_positions = np.random.uniform(lower, upper, size=(number_of_ballots, dim))
+
+        ballot_pool = []
```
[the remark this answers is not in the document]
cdonnay: [-1]*number_of_ballots

ballot_pool[i]=hjfsjkfkjfs

src/votekit/ballot_generator.py
```diff
@@ -1930,3 +1930,140 @@ def generate_profile(
...
+
+class ClusteredSpatial(BallotGenerator):
+    """
+    Clustered spatial model for ballot generation.
+    Assumes the candidates are uniformly distributed in d dimensional space.
+    Voters are then distributed normally around them, and vote with
+    ballots based on Euclidean distance to the candidates.
+
+    Args:
+        candidates (list): List of candidate strings.
```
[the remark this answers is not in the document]
cdonnay: also move distribution, distance to init method instead of generate_profile

src/votekit/elections/election_types.py
```diff
@@ -1117,3 +1118,183 @@ def __init__(
...
+            remaining = self.state.profile.get_candidates()
+
+            ballots = self.state.profile.ballots
+            weights = [b.weight for b in ballots]
+            random_ballot = random.choices(
+                self.state.profile.ballots, weights=weights, k=1
+            )[0]
+
+            # randomly choose a winner from the first place rankings
+            winning_candidate = list(random_ballot.ranking[0])[0]
```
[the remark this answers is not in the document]
cdonnay: random.choice(random_ballot.ranking[0])[0]

src/votekit/elections/election_types.py
```diff
@@ -1117,3 +1118,183 @@ def __init__(
...
+            u = random.uniform(0, 1)
+
+            if len(remaining) == 1:
+                winning_candidate = remaining[0]
+
+            elif u <= 1 / (len(remaining) - 1):
+                # Choose via proportional to squares
+                candidate_votes = {c: 0 for c in remaining}
+                for ballot in self.state.profile.get_ballots():
+                    top_choice = list(ballot.ranking[0])[0]
```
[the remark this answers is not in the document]
cdonnay: random.choice(random_ballot.ranking[0])[0]

src/votekit/elections/election_types.py
```diff
@@ -1117,3 +1118,183 @@ def __init__(
...
+                winning_candidate = np.random.choice(remaining, p=probabilities)
+
+            else:
+                ballots = self.state.profile.ballots
+                weights = [b.weight for b in ballots]
+                random_ballot = random.choices(
+                    self.state.profile.ballots, weights=weights, k=1
+                )[0]
+                # randomly choose a winner according to first place rankings
+                winning_candidate = list(random_ballot.ranking[0])[0]
```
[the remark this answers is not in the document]
cdonnay: random.choice(random_ballot.ranking[0])[0]

src/votekit/elections/election_types.py
```diff
@@ -1117,3 +1118,183 @@ def __init__(
...
+        return self.state
+
+
+class BoostedRandomDictator(RandomDictator):
+    """
+    Modified random dictator where we
+        - Choose a winner randomly from the distribution of first
+          place votes with probability (1 - 1/(# Candidates - 1))
+        - Choose a winner via a proportional to squares rule with
+          probability 1/(# of Candidates - 1)
```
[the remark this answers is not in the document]
cdonnay: instead of inheriting dictator, just do election

src/votekit/elections/election_types.py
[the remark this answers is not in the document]
kevin-q2: After some googling I believe list does not randomize the order of the set. Unfortunately random.choice() doesn't work on sets either. It could be slow but a good option might be: random.choice(list(random_ballot.ranking[0])). What do you think?

src/votekit/ballot_generator.py
```diff
@@ -1932,53 +1932,102 @@ def generate_profile(
             return pp
 
 
-class UniformSpatial(BallotGenerator):
+# Not sure where to put this! Poetry complains if I try to put it inside the class
```
cdonnay, line 1935: You can put this in the metrics/distances.py file.

src/votekit/ballot_generator.py
```diff
@@ -1932,53 +1932,102 @@ def generate_profile(
...
+    by an input distance function, randomly sample each voter's and
+    each candidate's positions from input voter and candidate distributions.
+    Using generate_profile() outputs a ranked profile which is consistent
+    with the sampled positions (respects distances).
 
     Args:
-        candidates (list): List of candidate strings.
-
+        candidates (list[str]): List of candidate strings.
+        voter_dist (Callable[..., np.ndarray]): Distribution to sample a single
```
cdonnay, line 1950: If there's a default, use (type, optional) in the doc string

cdonnay: If an arg/attr doc string is longer than one line, tab any line after the first.

src/votekit/ballot_generator.py
```diff
@@ -1932,53 +1932,102 @@ def generate_profile(
...
+        candidate_params: Optional[Dict[str, float]]: Parameters to be passed
+        to candidate_dist, defaults to uniform(0,1) in 2 dimensions.
+        distance: (Callable[[np.ndarray, np.ndarray], float]], optional):
+        Computes distance between a voter and a candidate,
+        defaults to euclidean distance.
     Attributes:
-        candidates (list): List of candidate strings.
+        candidates (list[str]): List of candidate strings.
+        voter_dist (Callable[..., np.ndarray]): Distribution to sample a single
+        voter's position from, defaults to uniform distribution.
```
cdonnay, line 1964: tab after first line

src/votekit/ballot_generator.py
```diff
@@ -1932,53 +1932,102 @@ def generate_profile(
...
+    with the sampled positions (respects distances).
 
     Args:
-        candidates (list): List of candidate strings.
-
+        candidates (list[str]): List of candidate strings.
+        voter_dist (Callable[..., np.ndarray]): Distribution to sample a single
+        voter's position from, defaults to uniform distribution.
+        voter_params: (dict[str, float], optional): Parameters to be passed to
+        voter_dist, defaults to uniform(0,1) in 2 dimensions.
```
cdonnay, line 1953: say defaults to None, which creates the unif(0,1) in 2D.

src/votekit/ballot_generator.py
```diff
@@ -1932,53 +1932,102 @@ def generate_profile(
...
+        candidate_dist: (Callable[..., np.ndarray]): Distribution to sample a
+        single candidate's position from, defaults to uniform distribution.
+        candidate_params: Optional[Dict[str, float]]: Parameters to be passed
```
cdonnay, line 1956: say defaults to None, which creates the unif(0,1) in 2D.

src/votekit/ballot_generator.py
cdonnay: Make this `Optional[dict[str, Any]]` in case users have more complicated distributions that need non-float parameters.

src/votekit/ballot_generator.py
cdonnay: Make this Optional[dict[str, Any]] in case users have more complicated distributions that need non-float parameters.

src/votekit/ballot_generator.py
cdonnay: same comments as in args.

src/votekit/ballot_generator.py
```diff
@@ -1932,53 +1932,102 @@ def generate_profile(
...
-            dim (int, optional): number of dimensions to use, defaults to 2d
-            lower (float, optional): lower bound for uniform distribution, defaults to 0
-            upper (float, optional): upper bound for uniform distribution, defaults to 1
-            seed (int, optional): seed for random generation
+            seed (int, optional): Seed for random generation, defaults to None
+            which resets the random seed.
 
         Returns:
-            Union[PreferenceProfile, Tuple]
+            preference profile, candidate positions, voter positions:
```
cdonnay, line 2015: Correct return string syntax is `Union[PreferenceProfile, Tuple]: explanation of what is returned`

cdonnay: @peterrrock2 , I'm just making a note that I think we can clean up how the `BallotGenerator` class is defined to avoid these weird type issues. The reason this is a Union[PrefProfile, Tuple] is b/c that's what the abstract method expects.

src/votekit/ballot_generator.py
```diff
@@ -1932,53 +1932,102 @@ def generate_profile(
...
+            [self.voter_dist(**self.voter_params) for v in range(number_of_ballots)]
+        )
 
         ballot_pool = []
-
-        for vp in voter_positions:
+        for v in voter_positions:
             distance_dict = {
-                c: np.linalg.norm(v - vp) for c, v, in candidate_position_dict.items()
+                i: self.distance(v, c) for i, c, in candidate_position_dict.items()
```
cdonnay, line 2030: This syntax is a bit confusing; save i for indices. I would do something like `cand_name, cand_coord`

src/votekit/ballot_generator.py
```diff
@@ -1993,68 +2042,106 @@ def generate_profile(
...
+class Clustered_DSpatial(BallotGenerator):
     """
-    Clustered spatial model for ballot generation.
-    Assumes the candidates are uniformly distributed in d dimensional space.
-    Voters are then distributed normally around them, and vote with
-    ballots based on Euclidean distance to the candidates.
+    Clustered spatial model for ballot generation. In some metric space
+    determined by an input distance function, randomly sample
+    each candidate's positions from input candidate distribution. Then
+    sample voters's positions from normal distributions centered around each
```
cdonnay, line 2050: Does it have to be normal?

src/votekit/ballot_generator.py
```diff
@@ -1993,68 +2042,106 @@ def generate_profile(
...
+    of the candidate's positions. Using generate_profile()
+    outputs a ranked profile which is consistent
+    with the sampled positions (respects distances).
 
     Args:
-        candidates (list): List of candidate strings.
-
+        candidates (list[str]): List of candidate strings.
+        voter_dist (Callable[..., np.ndarray]): Distribution to sample a single
```
cdonnay, line 2057: tabs for second lines

src/votekit/ballot_generator.py
```diff
@@ -1993,68 +2042,106 @@ def generate_profile(
...
+        voter's position from, defaults to uniform distribution.
+        voter_params: (dict[str, float], optional): Parameters to be passed to
```
cdonnay, line 2059: `Optional[dict[str,  Any]]`

src/votekit/ballot_generator.py
```diff
@@ -1993,68 +2042,106 @@ def generate_profile(
...
+        voter_dist, defaults to uniform(0,1) in 2 dimensions.
+        candidate_dist: (Callable[..., np.ndarray]): Distribution to sample a
+        single candidate's position from, defaults to uniform distribution.
+        candidate_params: Optional[Dict[str, float]]: Parameters to be passed
```
cdonnay, line 2063: `Optional[dict[str,  Any]]`

cdonnay: same comments as before, the default is None which causes behavior XYZ

src/votekit/ballot_generator.py
```diff
@@ -1993,68 +2042,106 @@ def generate_profile(
...
+        to candidate_dist, defaults to uniform(0,1) in 2 dimensions.
+        distance: (Callable[[np.ndarray, np.ndarray], float]], optional):
+        Computes distance between a voter and a candidate,
+        defaults to euclidean distance.
     Attributes:
-        candidates (list): List of candidate strings.
+        candidates (list[str]): List of candidate strings.
+        voter_dist (Callable[..., np.ndarray]): Distribution to sample a single
```
cdonnay, line 2070: tabs

cdonnay: all the same comments as args

src/votekit/ballot_generator.py
```diff
@@ -1993,68 +2042,106 @@ def generate_profile(
...
-        number_of_ballots: dict,
-        by_bloc: bool = False,
-        dim: int = 2,
-        lower: float = 0,
-        upper: float = 1,
-        std: float = 1,
-        seed: Optional[int] = None,
+        candidates: list[str],
+        voter_dist: Callable[..., np.ndarray] = np.random.uniform,
+        voter_params: Optional[Dict[str, np.ndarray]] = None,
```
cdonnay, line 2088: the dict should take [str, Any]

src/votekit/ballot_generator.py
```diff
@@ -1993,68 +2042,106 @@ def generate_profile(
...
+        candidate_dist: Callable[..., np.ndarray] = np.random.uniform,
+        candidate_params: Optional[Dict[str, float]] = None,
```
cdonnay, line 2090: the dict should take [str, Any]

src/votekit/ballot_generator.py
```diff
@@ -1993,68 +2042,106 @@ def generate_profile(
...
+            self.candidate_params = candidate_params
+
+        self.distance = distance
+
+    def generate_profile_with_dict(
+        self, number_of_ballots: dict, by_bloc: bool = False, seed: Optional[int] = None
     ) -> Union[PreferenceProfile, Tuple]:
         """
         Args:
             number_of_ballots (dict): The number of voters attributed
```
cdonnay, line 2115: Can you type set this? dict[str, int]

src/votekit/ballot_generator.py
```diff
@@ -1993,68 +2042,106 @@ def generate_profile(
...
-            upper (float, optional): upper bound for uniform distribution, defaults to 1
-            std (float, optional): standard deviation for voters normally distributed around
-                                   candidates, defaults to 1
-            seed (int, optional): seed for random generation
+            seed (int, optional): Seed for random generation, defaults to None
+            which resets the random seed.
 
         Returns:
-            Union[PreferenceProfile, Tuple]
+            preference profile, candidate positions, voter positions:
```
cdonnay, line 2122: same comment as before

src/votekit/ballot_generator.py
```diff
@@ -1993,68 +2042,106 @@ def generate_profile(
...
         voter_positions = []
         for c in self.candidates:
-            voter_positions.append(
-                np.random.normal(
-                    loc=candidate_position_dict[c],
-                    scale=std,
-                    size=(number_of_ballots[c], dim),
+            for v in range(number_of_ballots[c]):
+                voter_positions.append(
+                    np.random.normal(
```
cdonnay, line 2134: does this have to be the normal dist?

src/votekit/ballot_generator.py
```diff
@@ -1993,68 +2042,106 @@ def generate_profile(
...
 
         voter_positions_array = np.vstack(voter_positions)
 
         ballot_pool = []
-
-        for vp in voter_positions_array:
+        for vp in voter_positions:
             distance_dict = {
-                c: np.linalg.norm(v - vp) for c, v, in candidate_position_dict.items()
+                i: self.distance(vp, c) for i, c, in candidate_position_dict.items()
```
cdonnay, line 2144: same comment as before, confusing names

cdonnay (changes requested):
Looking good! Keep at it.

src/votekit/elections/election_types.py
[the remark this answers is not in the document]
peterrrock2: So the order of the set is determined by the hash of the underlying type. There is an environment variable, PYTHONHASHSEED, that controls this for security reasons. If you want this to be deterministic, you will need to use something other than a set. You can make use of the OrderedSet class in the collections module, or just use a dictionary. Convertiing to a list is also fine here. It doesn't really take up a lot of overhead in the grand scheme of things.

src/votekit/ballot_generator.py
```diff
@@ -1930,3 +1930,140 @@ def generate_profile(
...
+        ballot_pool = []
+
+        for vp in voter_positions:
+            distance_dict = {
+                c: np.linalg.norm(v - vp) for c, v, in candidate_position_dict.items()
+            }
+            candidate_order = sorted(distance_dict, key=distance_dict.__getitem__)
+            ballot_pool.append(candidate_order)
+
+        # reset the seed
```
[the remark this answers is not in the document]
peterrrock2: Yes, the seed setting is a global state modifier, so once it is set it affects everything in the active python instance forever until it is changed. Passing the `None` value here makes sure that the random numbers other parts of the program get are non-deterministic again

src/votekit/ballot_generator.py
```diff
@@ -1930,3 +1930,140 @@ def generate_profile(
...
+        ballot_pool = []
+
+        for vp in voter_positions_array:
+            distance_dict = {
+                c: np.linalg.norm(v - vp) for c, v, in candidate_position_dict.items()
+            }
+            candidate_order = sorted(distance_dict, key=distance_dict.__getitem__)
+            ballot_pool.append(candidate_order)
+
+        # reset the seed
```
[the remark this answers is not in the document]
peterrrock2: Yes. See my other comment

src/votekit/ballot_generator.py
```diff
@@ -1930,3 +1930,140 @@ def generate_profile(
...
+
+    Args:
+        candidates (list): List of candidate strings.
+
+    Attributes:
+        candidates (list): List of candidate strings.
+
+    """
+
+    def generate_profile_with_dict(
```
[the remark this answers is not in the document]
peterrrock2: I mean, you can just call it `generate_profile`. Abstract classes generally allow you to add additional keyword arguments in the derived class function. You just need to make sure that you at least include the keywords from the abstract class's original definition

src/votekit/elections/election_types.py
```diff
@@ -1310,3 +1312,125 @@ def run_election(self):
...
+            last_eliminated = None
+            for r in random_order:
+                # From each ballot, look at all the remaining candidates
+                # in A, and veto one which this ballot ranks lowest.
+                # This might be expensive, let me know if you have any ideas for how to improve
+                ballot = ballots[r]
+
+                max_rank = -1
+                max_rank_cands = []
+                for c in A.keys():
```
cdonnay, line 1378: If I'm understanding the election correctly, you want to choose a random candidate from the lowest ranking position. I would just access the last ranking position by `ballot.ranking[-1]` and skip this whole for loop.

cdonnay: The use of max is also confusing here since I think you want the "worst" ranked candidate. Max could mean "maximally ranked" which could mean a good candidate or something. I would just say "least_preferred" or something.

cdonnay (changes requested):

src/votekit/ballot_generator.py
kevin-q2: So it doesn't have to be normal, but it would have to be a distribution where the mean is a parameter. I kept it as normal because I thought that would often be the most natural choice. But I can definitely try add a more general form with some sort of check to make sure the input has a mean parameter.

src/votekit/elections/election_types.py
kevin-q2: So the trouble I had is that it's not always the very last ranking that I need to find. I keep a set of electable candidates called A from which candidates get eliminated as the process goes on. The loop finds each voters least preferred candidate among those in A. But yes this loop is not efficient and now I'm not sure its the right way to do multi-winner voting, so I'll have think about this and get back to you.

src/votekit/ballot_generator.py
```diff
@@ -2106,19 +2106,29 @@ class Clustered_DSpatial(BallotGenerator):
...
         if voter_params is None:
             # default params used for np.random.normal
-            self.voter_params = {"std": np.array(1.0), "size": np.array(2.0)}
+            self.voter_params = {"loc": 0, "std": np.array(1.0), "size": np.array(2.0)}
         else:
+            # Trying to ensure that the input distribution's mean is parameterized.
+            # This allows me to center it on the candidates themselves later
+            try:
+                voter_params['loc'] = 0
+                self.voter_dist(**voter_params)
```
kevin-q2, line 2127: I tried to fix Clustered_DSpatial so that it can take any general distribution for the voters. The only requirement it should satisfy is that the distribution has a 'loc' parameter specifying the mean of the distribution. Like in np.random.normal(loc = 0). Let me know if you have any thoughts or suggestions for how to test the input distribution for that.

src/votekit/elections/election_types.py
```diff
@@ -1117,3 +1120,317 @@ def __init__(
             seats=seats,
             tiebreak=tiebreak,
         )
+
+
+class RandomDictator(Election):
+    """
+    Choose a winner randomly from the distribution of first place votes. For multi-winner elections
+    repeat this process for every winner, removing that candidate from every voter's ballot
+    once they have been elected.
```
cdonnay, line 1129: For multi-winner, sample without replacement from FPV distribution?

src/votekit/elections/election_types.py
```diff
@@ -1117,3 +1120,317 @@ def __init__(
...
+
+    Args:
+      profile (PreferenceProfile): PreferenceProfile to run election on.
+      seats (int): Number of seats to elect.
+
+    Attributes:
+      _profile (PreferenceProfile): PreferenceProfile to run election on.
+      seats (int): Number of seats to be elected.
```
cdonnay, line 1137: m

src/votekit/elections/election_types.py
```diff
@@ -1117,3 +1120,317 @@ def __init__(
...
+        Determines if another round is needed.
+
+        Returns:
+            (bool) True if number of seats has not been met, False otherwise.
+        """
+        cands_elected = [len(s) for s in self.state.winners()]
+        return sum(cands_elected) < self.seats
+
+    def run_step(self):
+        if self.next_round():
```
cdonnay, line 1156: Describe in a doc string, remove most intermediate comments.

src/votekit/elections/election_types.py
```diff
@@ -1117,3 +1120,317 @@ def __init__(
...
+            self.state = ElectionState(
+                curr_round=self.state.curr_round + 1,
+                elected=elected,
+                eliminated_cands=[],
+                remaining=remaining,
+                profile=new_profile,
+                previous=self.state,
+            )
+
+            # if this is the last round, move remaining to eliminated
```
cdonnay, line 1189: Let's keep all the remaining candidates as remaining, since they were not technically eliminated in your election process.

src/votekit/elections/election_types.py
```diff
@@ -1117,3 +1120,317 @@ def __init__(
...
+        - With probability 1/(# of Candidates - 1)
+          Choose a winner via a proportional to squares rule.
+
+    For multi-winner elections
+    repeat this process for every winner, removing that candidate from every
+    voter's ballot once they have been elected.
+
+    Args:
+      profile (PreferenceProfile): PreferenceProfile to run election on.
+      seats (int): Number of seats to elect.
```
cdonnay, line 1223: m

src/votekit/elections/election_types.py
```diff
@@ -1117,3 +1120,317 @@ def __init__(
...
+        while self.next_round():
+            self.run_step()
+
+        return self.state
+
+
+class BoostedRandomDictator(Election):
+    """
+    Modified random dictator where
+        - With probability (1 - 1/(# Candidates - 1))
```
cdonnay, line 1212: n_cand?

src/votekit/elections/election_types.py
cdonnay, line 1: For all election types, raise ValueError on m for negative, 0, and bigger than number of candidates.

src/votekit/elections/election_types.py
```diff
@@ -1117,3 +1120,317 @@ def __init__(
...
+        Determines if another round is needed.
+
+        Returns:
+            (bool) True if number of seats has not been met, False otherwise.
+        """
+        cands_elected = [len(s) for s in self.state.winners()]
+        return sum(cands_elected) < self.seats
+
+    def run_step(self):
+        if self.next_round():
```
cdonnay, line 1246: doc string

src/votekit/elections/election_types.py
```diff
@@ -1117,3 +1120,317 @@ def __init__(
...
+            remaining = self.state.profile.get_candidates()
+            u = random.uniform(0, 1)
+
+            if len(remaining) == 1:
+                winning_candidate = remaining[0]
+
+            elif u <= 1 / (len(remaining) - 1):
+                # Choose via proportional to squares
+                candidate_votes = {c: 0 for c in remaining}
```
cdonnay, line 1255: Helper function: first_place_votes in utils module.

src/votekit/elections/election_types.py
```diff
@@ -1117,3 +1120,317 @@ def __init__(
...
+                for ballot in self.state.profile.get_ballots():
+                    top_choice = random.choice(list(ballot.ranking[0]))
+                    candidate_votes[top_choice] += float(ballot.weight)
+
+                squares = np.array(
+                    [(i / len(remaining)) ** 2 for i in candidate_votes.values()]
```
cdonnay, line 1261: Divide by total ballot weight which is the total number of voters.

src/votekit/elections/election_types.py
```diff
@@ -1117,3 +1120,317 @@ def __init__(
...
+                curr_round=self.state.curr_round + 1,
+                elected=elected,
+                eliminated_cands=[],
+                remaining=remaining,
+                profile=new_profile,
+                previous=self.state,
+            )
+
+            # if this is the last round, move remaining to eliminated
+            if not self.next_round():
```
cdonnay, line 1298: Delete this block

src/votekit/elections/election_types.py
```diff
@@ -1117,3 +1120,317 @@ def __init__(
...
+class PluralityVeto(Election):
+    """
+    Scores each candidate by their plurality (number of first) place votes,
+    then in a randomized order it lets each voter decrement the score of their
+    least favorite candidate. The candidate with the largest score at the end is
+    then chosen as the winner.
+
+    Args:
+      profile (PreferenceProfile): PreferenceProfile to run election on.
+      seats (int): Number of seats to elect.
```
cdonnay, line 1326: m

src/votekit/elections/election_types.py
```diff
@@ -1117,3 +1120,317 @@ def __init__(
...
+        Determines if another round is needed.
+
+        Returns:
+            (bool) True if number of seats has not been met, False otherwise.
+        """
+        cands_elected = [len(s) for s in self.state.winners()]
+        return sum(cands_elected) < self.seats
+
+    def run_step(self):
+        if self.next_round():
```
cdonnay, line 1349: doc string

src/votekit/elections/election_types.py
```diff
@@ -1117,3 +1120,317 @@ def __init__(
...
+            candidates = self.state.profile.get_candidates()
+            ballots = self.state.profile.ballots
+
+            # First, count a plurality score for each of the candidates
+            candidate_approvals = {c: Fraction(0) for c in candidates}
```
cdonnay, line 1354: use first_place_votes

src/votekit/elections/election_types.py
```diff
@@ -1117,3 +1120,317 @@ def __init__(
...
+        """
+        Determines if another round is needed.
+
+        Returns:
+            (bool) True if number of seats has not been met, False otherwise.
+        """
+        cands_elected = [len(s) for s in self.state.winners()]
+        return sum(cands_elected) < self.seats
+
+    def run_step(self):
```
cdonnay, line 1348: Okay: let's refactor this. Each round should end when one candidate is eliminated.
Then remove them from the profile.  Start a new round.

You'll need to ensure that the random order is preserved. Should each ballot have weight 1? Yes: de-coupling the ballots with weight bigger than 1. In the init method, validate that each weight is an integer.

src/votekit/metrics/distances.py
```diff
@@ -145,3 +145,7 @@ def em_array(pp: PreferenceProfile) -> list:
     ]
 
     return elect_distr
+
+
+def euclidean_dist(point1: np.ndarray, point2: np.ndarray) -> float:
+    return float(np.linalg.norm(point1 - point2))
```
cdonnay, line 151: Add `order=2` to make sure norm is the L2 norm.

cdonnay (changes requested):
Going great, Kevin!

src/votekit/ballot_generator.py
```diff
@@ -1948,3 +1949,244 @@ def generate_profile(
...
+    ):
+        super().__init__(candidates=candidates)
+        self.voter_dist = voter_dist
+        self.candidate_dist = candidate_dist
+
+        if voter_params is None:
+            # default params used for np.random.uniform
+            self.voter_params = {"low": 0.0, "high": 1.0, "size": 2.0}
+        else:
+            self.voter_params = voter_params
```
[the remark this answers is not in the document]
kevin-q2: It seems that inspect.signature() does not work for some functions written with Cython. And unfortunately this means I cannot use it to check the parameters for np.random.uniform, np.random.normal, etc. My proposed solution is to include some try/except statements to see if the input parameters actually work with the input distribution.

src/votekit/elections/election_types.py
kevin-q2: I'm going to leave this as is for now and take some more time to think about a fix

PR #134, opened 2024-07-22
ElectionState

cdonnay:
Modify the `ElectionState` to a dataclass. Remove all methods which are now handled by `Election`.


peterrrock2 (approved):
Looks good!

PR #135, opened 2024-07-23
edit Election class

cdonnay:
Edit the existing Election class. 
- Move methods that used to be in ElectionState to Election class.
- add tests


src/votekit/models.py
```diff
@@ -1,95 +1,292 @@
...
+        Fetches the profile and ElectionState of the given round number.
+
+        Args:
+            round_number (int, optional): The round number to fetch. Defaults to -1,
+                which fetches the final round.
+
+        Returns:
+            tuple[PreferenceProfile, ElectionState]
+        """
+        if round_number < -1:
```
peterrrock2, line 85: Let `get_profile` raise the error for you here rather than handling it at this level. This function can literally be 

```
return (self.get_profile(round_number, self.election_states[round_number])
```

if we change the `get_profile` method to work like a list in terms of indexing, it would also make this function look/feel more natural

src/votekit/models.py
```diff
@@ -1,95 +1,292 @@
...
+        Args:
+            round_number (int, optional): Round number, defaults to -1 which corresponds to
+                the final round.
+
+        Returns:
+            list[frozenset[str]]: List of winning candidates in order of election. Candidates
+                in the same set were elected simultaneously, i.e. in the final ranking
+                they are tied.
+        """
+        if round_number < -1:
```
peterrrock2, line 106: You don't need this value error since the list type will take care of erroneous indexing for you

src/votekit/models.py
```diff
@@ -1,95 +1,292 @@
...
+            round_number (int, optional): Round number, defaults to -1 which corresponds to
+                the final round.
 
-        self.state = ElectionState(curr_round=0, profile=self._profile)
+        Returns:
+            list[frozenset[str]]: List of eliminated candidates in reverse order of elimination.
+                Candidates in the same set were eliminated simultaneously, i.e. in the final ranking
+                they are tied.
+        """
+        if round_number < -1:
```
peterrrock2, line 132: Again, let the list handle the error here

peterrrock2: If you want to make your own error message, you can catch the IndexError that the list raises, and then raise a new index error with a custom message from that

src/votekit/models.py
```diff
@@ -1,95 +1,292 @@
...
+
+        Args:
+            round_number (int, optional): Round number, defaults to -1 which corresponds to
+                the final round.
+
+        Returns:
+            list[frozenset[str]]: List of sets of remaining candidates. Ordering of tuple
+            denotes ranking of remaining candidates, sets denote ties.
         """
+        if round_number < -1:
```
peterrrock2, line 158: Use the list

src/votekit/models.py
```diff
@@ -1,95 +1,292 @@
...
+                the final round.
 
         Returns:
-            The ElectionState object for round = step.
+            list[set[str]]: Ranking of candidates.
         """
-        if isinstance(step, int):
-            if step < 0:
-                raise ValueError("Step must be a non-negative integer.")
+        if round_number < -1:
```
peterrrock2, line 177: Here the `get_*` functions can handle the error

src/votekit/models.py
```diff
@@ -1,95 +1,292 @@
...
+
+        Args:
+            round_number (int, optional): Round number, defaults to -1 which corresponds to
+                the final round.
+        Returns:
+            pd.DataFrame:
+                Data frame displaying candidate, status (elected, eliminated,
+                remaining), and the round their status updated.
+        """
+        if round_number < -1:
```
peterrrock2, line 205: Here you can do the same modular indexing thing that I describe in the `get_profile` function

src/votekit/models.py
```diff
@@ -1,95 +1,292 @@
...
+                `self._run_election()`. Defaults to False.
+
+        Returns:
+            PreferenceProfile: The profile of ballots after the round is completed.
+        """
         pass
 
     @abstractmethod
-    def run_election(self, *args: Any, **kwargs: Any):
+    def _next_round(self) -> bool:
```
peterrrock2, line 260: It would be good to change the name of this. The name `_next_round` makes me think that it returns the next round of the election. Maybe something like `_is_finished` would be better (of course, this name suggests that you have to invert the logic for things)

src/votekit/models.py
```diff
@@ -1,95 +1,292 @@
...
-            else:
-                num_ties = 0
-                for rank in ballot.ranking:
-                    if len(rank) > 1:
-                        num_ties += 1
+        self.election_states.append(ElectionState(remaining=remaining, scores=scores))
 
-                resolved_ties = fix_ties(ballot)
-                new_ballots += recursively_fix_ties(resolved_ties, num_ties)
+        while self._next_round():
```
peterrrock2, line 286: Yeah, this would read better as `while not self._is_finished():`

src/votekit/models.py
```diff
@@ -1,95 +1,292 @@
...
+
+        Returns:
+            PreferenceProfile
+
+        """
+        if round_number < -1:
+            raise ValueError("round_number must be -1 or non-negative.")
+
+        if round_number == -1:
+            round_number = len(self)
```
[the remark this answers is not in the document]
peterrrock2: This likely doesn't cause an issue now since an implementation of `_run_step` should do nothing if the election is over, but let's not rely on that always being the case and be intentional with our indexing so that we don't accidentally end up with undefined behaviour

tests/elections/test_electon_model.py
peterrrock2, line 1: These are looking good so far!

peterrrock2 (changes requested):

tests/elections/test_electon_model.py
cdonnay: :)

src/votekit/models.py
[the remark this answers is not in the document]
cdonnay: This should be okay actually; the len(self) in a 3 round election is 3, so the function runs 3 times.

cdonnay:
@peterrrock2 , I addressed your comments.

peterrrock2 (approved):
Looks good to me!

PR #136, opened 2024-07-24
edit utils

cdonnay:
Edit and clean the utils module: 

- remove functions that are not used in codebase
- consolidate several scoring functions to use a general scoring framework from a score vector


peterrrock2 (approved):
Looking good! Also, thank you for the thorough testing on this!

PR #138, opened 2024-08-06
Ballot

cdonnay:
- Add scores parameter to ballots, adapt eq and str methods
- Create validator methods to handle converting ints/floats to Fraction
- Tests

### DIFF
--- a/tests/test_ballot.py
+++ b/tests/test_ballot.py
@@ -1,0 +1,83 @@
+from votekit import Ballot
+from fractions import Fraction
+import pytest
+
+
+def test_ballot_init():
+    b = Ballot()
+    assert not b.ranking
+    assert b.weight == Fraction(1)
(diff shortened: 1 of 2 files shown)


tests/test_ballot.py
```diff
@@ -0,0 +1,83 @@
...
+    )
+    assert b != Ballot(
+        ranking=[{"A"}, {"B"}, {"C"}],
+        weight=3,
+        voter_set={"Chris"},
+        id="91010",
+    )
+
+
+def test_ballot_str():
```
peterrrock2, line 73: This test doesn't look like it is finished

peterrrock2 (changes requested):
It looks like there is one unfinished test in this PR

peterrrock2 (approved):

PR #139, opened 2024-08-06
PreferenceProfile

cdonnay:
- Freeze PreferenceProfile
- Adapt to score ballots
- Use validator methods


peterrrock2 (changes requested):
Some small tweaks again!

peterrrock2 (approved):

PR #141, opened 2024-08-06
Election model

cdonnay:
- validate profile method 
- modular indexing support for get methods
-  tests


peterrrock2 (approved):
Looks good!

PR #145, opened 2024-08-08
clean ups from 3.0.0 edits

cdonnay:
Changes in syntax to src and tests required by changes to PrefProfile, elections, utils, and ballots.


tests/test_e2e.py
```diff
@@ -102,15 +96,15 @@ def test_generate_election_diff_res():
...
-    outcome_irv = election_irv.run_election().winners()
-    outcome_plurality = election_plurality.run_election().winners()
-    outcome_seq = election_seq.run_election().winners()
+    outcome_borda = election_borda.get_elected()
+    outcome_irv = election_irv.get_elected()
+    outcome_plurality = election_plurality.get_elected()
+    outcome_seq = election_seq.get_elected()
     # outcome_sntv = election_sntv.run_election().get_all_winners()
 
     print(outcome_borda)
```
peterrrock2, line 110: Why do the commented lines and print statements exist in this?

peterrrock2 (approved):
One small thing that you can take or leave in here. Go ahead and merge if / when you make the changes

tests/test_e2e.py
cdonnay: No idea.

PR #157, opened 2025-01-06
Contributing guidelines

cdonnay:
- Add contributing guidelines to documentation. 
- Update the Readme to reflect this. 
- Add instructions for GitHub desktop users.
- Update poetry dependencies for developers, specifically adding sphinx dependencies for documentation.


docs/package_info/contributing.rst
```diff
@@ -0,0 +1,147 @@
...
+
+**Writing Tests**: If you write a new feature, please make sure that it is
+included in a test somewhere. A test is a short piece of code that checks that the package is working as intended.
+While it is not feasible to test every single aspect of a package, adding tests for new features is a crucial part
+of open-source software. You can use our `current tests <https://github.com/mggg/VoteKit/blob/main/tests>`_  as a starting place 
+to show you how these work. Depending on the scale of your feature, you may need multiple tests.
+At minimum, your tests should 
+
+- check the basic functionality of your feature, such as a test case of an algorithm or an instantiation of a class, and 
+- ensure that any error messages are correctly raised.
```
peterrrock2, line 26: As a part of the error messages, it is expected that you try to predict, test for, and handle relevant edge cases. It will not be possible to get all of them, but obvious edge cases should be checked (type errors, empty elements, etc.)

docs/package_info/contributing.rst
```diff
@@ -0,0 +1,147 @@
...
+
+**Documentation**: For any new features that you add, please make sure to include
+a comprehensive docstring. We have a defined format for docstrings that we use
+throughout the `codebase <https://github.com/mggg/VoteKit/blob/main/src/votekit>`_, so please make sure that any additions are consistent
+with that format.
```
peterrrock2, line 31: You should state what style we use google style outlined here:

https://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings

However, we move the first line of the docstring onto its own line.

peterrrock2 (changes requested):
Looking good! A couple of small notes here that we should address before pushing this out

docs/package_info/contributing.rst
peterrrock2: Also we should probably give a template for people that can't be bothered to look

cdonnay:
@peterrrock2 , addressed both of your comments. Thanks!!

peterrrock2 (approved):
Looks good!

PR #174, opened 2025-01-09
Fix default Borda scoring

cdonnay:
The old Borda scoring function assumed that the length of a ballot was the number of candidates in the profile. We now allow users to choose what the maximum Borda score is, and make the default that it checks the PrefProfile for the maximum ballot length.

- we added a max_ballot_length parameter to the PrefProfile object, which defaults to the longest observed ballot.


tests/test_utils.py
peterrrock2, line 1: We should probably have tests that account for some weird misuse cases. For example, having borda_max=50 when there are only 3 candidates or borda_max = 1 when there are ballots of length 5.

cdonnay:
@peterrrock2 this is ready for another look! I added the tests you suggested, and also allowed users to choose between three different scoring conventions for vector based scoring in the presence of ties.

peterrrock2 (approved):
Overall looking good! Small note: we should spell out in each of the doc strings that "low" makes every candidate not marked tie for last which is usually given a score of 0 (I do not think that this is the most intuitive behaviour if you are unfamiliar with the different versions of borda)

cdonnay:
The default behavior was actually changed with this PR; unlisted candidates are no longer listed as tied for last (although you can use another helper function in `utils` to do that if you want). In other words, these scoring functions only see exactly the ballots you give them, so making unlisted candidates tied for last is moved outside these functions. Now, any unlisted candidate just gets 0 points regardless of tie convention.

PR #178, opened 2025-02-11
Comentions Matrix

cdonnay:
Add comentions matrix to codebase. There are two other matrices (distance, boost) that will be added, so any feedback on structure now will save me some pain down the line!


src/votekit/matrices/mentions.py
```diff
@@ -0,0 +1,116 @@
...
+    """
+
+    cands: set[str] = set()
+    if ballot.scores:
+        cands = cands.union(ballot.scores.keys())
+
+    if ballot.ranking:
+        cands = cands.union(c for s in ballot.ranking for c in s)
+
+    return i in cands and j in cands
```
peterrrock2, line 51: I think that you can do this in one line

```suggestion
def comention(i: str, j: str, ballot: Ballot) -> bool:
    """
    Takes candidates i,j and returns true if they both appear on the ballot, either in the ranking
    or the scoring. Candidates who receive 0 points are not counted as mentioned.

    Args:
      i (str): Candidate name.
      j (str): Candidate name.
      ballot (Ballot): Ballot.

    Returns:
      bool: True if both i and j appear in ballot.
    """
    return mention(i, ballot) and mention(j, ballot)
```

src/votekit/matrices/mentions.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+def comentions_matrix(
+    pref_profile: PreferenceProfile, candidates: list[str], symmetric: bool = False
+) -> np.ndarray:
+    """
+    Takes a preference profile and converts to a numpy array
+    where the i,j entry shows the number of times candidates i,j were mentioned on the same
+    ballot with i above j. There is an option to make it symmetric so that the i,j entry is just
+    the comentions of i,j. Comentions are counted using ballot weight. The indexing of the
```
peterrrock2, line 89:
```suggestion
    the number of comentions of i and j. Comentions are counted using ballot weight. The indexing of the
```

src/votekit/matrices/mentions.py
```diff
@@ -41,19 +46,12 @@ def comention(i: str, j: str, ballot: Ballot) -> bool:
...
 
-    cands: set[str] = set()
-    if ballot.scores:
-        cands = cands.union(ballot.scores.keys())
-
-    if ballot.ranking:
-        cands = cands.union(c for s in ballot.ranking for c in s)
-
-    return i in cands and j in cands
+    return mention(i, ballot) and mention(j, ballot)
```
peterrrock2, line 49: You can now do this with just 

```
return mention([i,j], ballot)
```

this will also be faster. Or you can just get rid of the `comention` function

peterrrock2 (changes requested):

src/votekit/matrices/mentions.py
cdonnay: Renamed mentions to comentions and then just will use it with one candidate for mentions!

peterrrock2 (approved):

PR #179, opened 2025-02-12
Add boost and candidate distance matrix

cdonnay:
Add boost and candidate distance matrix.

### DIFF
--- a/src/votekit/matrices/boost.py
+++ b/src/votekit/matrices/boost.py
@@ -1,0 +1,75 @@
+from ..pref_profile import PreferenceProfile
+from .comentions import comention
+import numpy as np
+from typing import Tuple
+from ._utils import _convert_dict_to_matrix
+from fractions import Fraction
+
+
+def boost_prob(i: str, j: str, pref_profile: PreferenceProfile) -> Tuple[float, float]:
+    """
+    Takes candidates i,j and a preference profile and computes the conditional
+    P(mention i | mention j) and P(mention i). If candidate j is never mentioned,
+    or if the profile has 0 total ballot weight, it returns
+    numpy.nan for the respective probability.
+
+    Args:
+      i (str): Candidate.
+      j (str): Candidate.
+      pref_profile (PreferenceProfile): Profile.
+
+    Returns:
+      tuple[float, float]: P(mention i | mention j), P(mention i)
+    """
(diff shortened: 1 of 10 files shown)


src/votekit/matrices/boost.py
```diff
@@ -0,0 +1,75 @@
...
+            i_mentions += ballot.weight
+
+        if comention(j, ballot):
+            j_mentions += ballot.weight
+
+        if comention([i, j], ballot):
+            both_mentions += ballot.weight
+
+    return (
+        float(both_mentions) / j_mentions if j_mentions else np.nan,
```
peterrrock2, line 40: Let's just be very explicit that we are checking that `j_mentions` is not 0

```suggestion
        float(both_mentions) / j_mentions if (j_mentions != 0) else np.nan,
```

src/votekit/matrices/boost.py
```diff
@@ -0,0 +1,75 @@
...
+        (
+            float(i_mentions) / pref_profile.total_ballot_wt
+            if pref_profile.total_ballot_wt
```
peterrrock2, line 43: Again, let's be explicit that we are checking for 0 denominators here

```suggestion
            if (pref_profile.total_ballot_wt != 0)
```

src/votekit/matrices/distance.py
```diff
@@ -0,0 +1,85 @@
...
+                dist_matrix[i][j] += d * ballot.weight
+                weight_matrix[i][j] += ballot.weight
+
+            # i < j
+            elif d < 0:
+                dist_matrix[j][i] += (-d) * ballot.weight
+                weight_matrix[j][i] += ballot.weight
+
+            elif np.isnan(d):
+                continue
```
peterrrock2, line 77: Get rid of  the last `elif` since we are at the end of the loop anyway

```suggestion
            if d >= 0:
                dist_matrix[i][j] += d * ballot.weight
                weight_matrix[i][j] += ballot.weight

            # i < j
            elif d < 0:
                dist_matrix[j][i] += (-d) * ballot.weight
                weight_matrix[j][i] += ballot.weight
```

peterrrock2 (changes requested):

peterrrock2 (approved):

PR #180, opened 2025-02-13
Add matrix heatmap plotting

cdonnay:
Add matrix heatmap plotting code.


src/votekit/matrices/heatmap.py
peterrrock2, line 1: Lol, I think that this looks good, but most of this is my code, so I don't know if that counts!

peterrrock2 (approved):

PR #185, opened 2025-02-20
Profile stat wrapper for bar plot

cdonnay:
Add a wrapper to multi bar plot that allows users to input a profile and a statistic, predefined or defined by the user.


peterrrock2 (approved):
Lookin' good!

PR #203, opened 2025-03-19
Update the condense_ballot() method

cdonnay:
- Update the condense_ballots() method of PreferenceProfile to also include the voter_set attribute.
- Rename condense_ballots to group_ballots to avoid confusion with new terminology of condense ballot introduced in Portland report. (That now refers to removing empty ranking positions and moving lower candidates up ballot).
- Update other functions in code base to include voter_set and id attribute

### DIFF
--- a/src/votekit/pref_profile.py
+++ b/src/votekit/pref_profile.py
@@ -362,7 +362,7 @@ class PreferenceProfile:
     # set repr to print outputs
     __repr__ = __str__
 
-    def condense_ballots(self) -> PreferenceProfile:
+    def group_ballots(self) -> PreferenceProfile:
         """
         Groups ballots by rankings and scores and updates weights. Retains voter sets, but
         loses ballot ids.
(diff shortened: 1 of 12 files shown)


src/votekit/pref_profile.py
```diff
@@ -362,49 +362,63 @@ def __str__(self) -> str:
...
+
+        for i, (ballot, ballot_dict) in enumerate(seen_ballots.items()):
+            new_ballots[i] = Ballot(
+                ranking=ballot.ranking,
+                scores=ballot.scores,
+                weight=ballot_dict["weight"],  # type: ignore[arg-type]
+                voter_set=(
+                    ballot_dict["voter_set"]  # type: ignore[arg-type]
+                    if len(ballot_dict["voter_set"])  # type: ignore[arg-type]
+                    else None
```
peterrrock2, line 407:
```suggestion
                voter_set=(
                    ballot_dict.get("voter_set", None)  # type: ignore[arg-type]
```

peterrrock2: I assume here that an empty voter set and None are to to be treated the same. Maybe we want to think about having the default value be the empty set so that we don't need to deal with 'None' value issues

peterrrock2 (changes requested):

src/votekit/pref_profile.py
cdonnay: Yes, we should change the default to be the empty set!

peterrrock2 (approved):

PR #204, opened 2025-03-21
Cleaning module

cdonnay:
An attempt to clean the cleaning module

- all cleaning functions are now based on `clean_profile` which applies a function to the list of ballots
- parameters added to track the number of affected ballots, as well as various profile attributes like candidate list and max ballot length
- elections now check that they have enough candidates with votes to end
- mentions score dict is now the only one that includes 0 vote candidates.


src/votekit/cleaning.py
```diff
@@ -1,185 +1,273 @@
...
-    pp: PreferenceProfile, keep_candidates: bool = False
-) -> PreferenceProfile:
+def clean_profile(
+    profile: PreferenceProfile,
+    clean_ballot_func: Callable[[Ballot], Ballot],
+    remove_empty_ballots: bool = True,
+    remove_zero_weight_ballots: bool = True,
+    retain_original_candidate_list: bool = True,
+    retain_original_max_ballot_length: bool = True,
+) -> CleanedProfile:
```
peterrrock2, line 16: We should talk about if this is how we would like for multiple rounds of cleaning to behave

peterrrock2 (changes requested):
You might want to change instances of PreferenceProfile to CleanProfile when instantiating an election type. So, I am thinking that you make a raw preference profile out of the ballots that are passed, and then you apply the cleaning function. That way, when users go to use the election method, they can clearly see that the election was run on potentially altered data. This would also be a good thing to show off in a tutorial somewhere

src/votekit/cleaning.py
```diff
@@ -264,10 +270,141 @@ def remove_cand(
...
+    remove_empty_ballots: bool = True,
+    remove_zero_weight_ballots: bool = True,
+    retain_original_candidate_list: bool = True,
+    retain_original_max_ranking_length: bool = True,
+) -> CleanedProfile:
+    """
+    Given a profile, removes any empty frozensets from the ballot rankings and condenses the
+    resulting ranking. If a ranking only has trailing empty positions, the condensed ballot is
+    considered equivalent. For example, (A,B,{},{}) is mapped to (A,B) but considered unaltered
+    since the ranking did not change.
```
peterrrock2, line 351: Make it clear that this function and `remove_repeated_candidates` call `clean_profile` with a particular cleaning rule, and which keeps track of some extra index stuff.

peterrrock2: Probably need a `remove_and_condense`  method for STV

src/votekit/elections/election_types/ranking/alaska.py
```diff
@@ -144,9 +144,11 @@ def _run_step(
...
             eliminated = plurality.get_remaining()
             tiebreaks = plurality.election_states[-1].tiebreaks
-            new_profile: PreferenceProfile = remove_cand(
-                [c for s in eliminated for c in s],
-                profile,
+            new_profile: PreferenceProfile = condense_profile(
+                remove_cand(
+                    [c for s in eliminated for c in s],
+                    profile,
+                )
```
peterrrock2, line 151: Change over to a single method?

peterrrock2 (commented):
Add a new method called `remove_and_condense` that replaces the lines


```
        new_profile = condense_profile(
            remove_cand([c for s in elected for c in s], profile)
        )
```

wherever they appear. This will speed up things by making it so that we do not iterate through the profile twice.

src/votekit/pref_profile/cleaned_pref_profile.py
```diff
@@ -0,0 +1,162 @@
...
+        ballots (tuple[Ballot]): Tuple of ``Ballot`` objects.
+        candidates (tuple[str]): Tuple of candidate strings.
+        max_ranking_length (int): The length of the longest allowable ballot, i.e., how
+            many candidates are allowed to be ranked in an election.
+        df (pandas.DataFrame): Data frame view of the ballots.
+        candidates_cast (tuple[str]): Tuple of candidates who appear on any ballot with positive
+            weight, either in the ranking or in the score dictionary.
+        total_ballot_wt (Fraction): Sum of ballot weights.
+        num_ballots (int): Length of ballot list.
+    """
```
peterrrock2, line 50:
```suggestion
    """
```

src/votekit/pref_profile/cleaned_pref_profile.py
```diff
@@ -0,0 +1,162 @@
...
+
+    parent_profile: SkipValidation[PreferenceProfile | CleanedProfile] = (
+        PreferenceProfile()
+    )
+    df_index_column: list[int] = field(default_factory=list)
+    no_weight_altr_ballot_indices: set[int] = field(default_factory=set)
+    no_ranking_and_no_scores_altr_ballot_indices: set[int] = field(default_factory=set)
```
peterrrock2, line 57:
```suggestion
    no_rank_no_score_altr_idxs: set[int] = field(default_factory=set)
```

src/votekit/pref_profile/cleaned_pref_profile.py
```diff
@@ -0,0 +1,162 @@
...
+            weight, either in the ranking or in the score dictionary.
+        total_ballot_wt (Fraction): Sum of ballot weights.
+        num_ballots (int): Length of ballot list.
+    """
+
+    parent_profile: SkipValidation[PreferenceProfile | CleanedProfile] = (
+        PreferenceProfile()
+    )
+    df_index_column: list[int] = field(default_factory=list)
+    no_weight_altr_ballot_indices: set[int] = field(default_factory=set)
```
peterrrock2, line 56: Shorten names

src/votekit/pref_profile/cleaned_pref_profile.py
```diff
@@ -0,0 +1,162 @@
...
+                )
+            )
+
+        if set().union(*index_sets) != set(self.parent_profile.df.index):
+            raise ValueError(
+                (
+                    "Union of ballot indices must equal the parent profile df index "
+                    "column."
+                )
+            )
```
peterrrock2, line 122: Print out the set minus in the error

peterrrock2 (approved):
Looking good! Thanks for all of your work on this!

PR #211, opened 2025-05-13
3.2.0 stv speed: PrefProfile from df

cdonnay:
This first pass at improving the speed of the STV election is focused on preferenceprofiles. Previously a profile had to be constructed from a list of ballots, and the conversion between the list and df structures was costly. This PR 
- allows a profile to be initialized from a properly formatted dataframe instead of a ballot list 
- caches the ballot property of profiles so that if constructed from a df, only need to do the conversion if called (which does NOT occur in STV, thus saving us speed)

Todo: implement cleaning methods that use the df structure.


src/votekit/pref_profile/pref_profile.py
```diff
@@ -598,7 +597,10 @@ def _validate_csv_format(csv_data: list[list[str]]):
     _validate_csv_ballot_rows(csv_data)
 
 
-@dataclass(frozen=True, config=ConfigDict(arbitrary_types_allowed=True))
+@dataclass(
+    frozen=True,
+    config=ConfigDict(arbitrary_types_allowed=True),
+)
```
peterrrock2, line 603: Get rid of the dataclass and write custom __setattr__ and __delattr__ methods

(c.f. [Docs](https://docs.python.org/3/library/dataclasses.html#dataclasses.dataclass))

src/votekit/pref_profile/pref_profile.py
```diff
@@ -663,6 +666,29 @@ class PreferenceProfile:
...
+        if not self.df.equals(pd.DataFrame()) and ballots != tuple():
+            raise ProfileError(
+                "Cannot pass a dataframe and a ballot list to profile init method. Must pick one."
+            )
+
+        elif self.df.equals(pd.DataFrame()) and ballots != tuple():
+            self._ballot_list_to_df(ballots)
+
+        elif not self.df.equals(pd.DataFrame()) and ballots == tuple():
+            self._set_class_attrs_from_init_df()
```
peterrrock2, line 686:
```suggestion
        if not self.df.equals(pd.DataFrame()) and ballots != tuple():
            raise ProfileError(
                "Cannot pass a dataframe and a ballot list to profile init method. Must pick one."
            )

        elif self.df.equals(pd.DataFrame()):
            self._ballot_list_to_df(ballots)

        else:
            self._set_class_attrs_from_init_df()
```

src/votekit/pref_profile/pref_profile.py
```diff
@@ -685,17 +711,16 @@ def cands_must_be_unique_strip_whitespace(
 
         return tuple([c.strip() for c in candidates])
 
-    @model_validator(mode="after")
-    def cands_not_ranking_columns(
+    def _cands_not_ranking_columns(
         self,
-    ) -> Self:
+    ):
```
peterrrock2, line 716:
```suggestion
    ) -> None:
```

src/votekit/pref_profile/pref_profile.py
```diff
@@ -901,13 +930,13 @@ def __init_formatted_df(
...
 
-    def __set_class_attrs_from_df(
+    def __set_class_attrs_from_constructed_df(
         self,
         df: pd.DataFrame,
         candidates_cast: list[str],
         contains_rankings_indicator: bool,
         contains_scores_indicator: bool,
-    ) -> Self:
+    ):
```
peterrrock2, line 939:
```suggestion
    ) -> None:
```

src/votekit/pref_profile/pref_profile.py
```diff
@@ -938,25 +967,25 @@ def __set_class_attrs_from_df(
                 "contains_scores is True but we found no ballots with scores."
             )
 
-        return self
-
-    @model_validator(mode="after")
-    def ballot_list_to_df(self) -> Self:
+    def _ballot_list_to_df(self, ballots: tuple[Ballot, ...]):
```
peterrrock2, line 970:
```suggestion
    def _save_ballot_list_to_internal_df(self, ballots: tuple[Ballot, ...]) -> None:
```

Maybe consider doing this without side-effects?

src/votekit/pref_profile/pref_profile.py
peterrrock2, line 1: None return type annotations

peterrrock2: General note: Remove the Dataclass and do it custom

peterrrock2 (changes requested):

src/votekit/pref_profile/pref_profile.py
```diff
@@ -653,59 +650,49 @@ class PreferenceProfile:
...
-    contains_scores: Optional[bool] = None
-
-    @field_validator("candidates")
-    @classmethod
-    def cands_must_be_unique_strip_whitespace(
-        cls, candidates: tuple[str, ...]
-    ) -> tuple[str, ...]:
-        """
-        Checks that candidate names are unique and strips trailing and leading whitespace.
+    _is_frozen = False
```
peterrrock2, line 653:
```suggestion
    _is_frozen: bool = False
```

src/votekit/pref_profile/pref_profile.py
```diff
@@ -598,7 +595,6 @@ def _validate_csv_format(csv_data: list[list[str]]):
     _validate_csv_ballot_rows(csv_data)
 
 
-@dataclass(frozen=True, config=ConfigDict(arbitrary_types_allowed=True))
 class PreferenceProfile:
```
peterrrock2, line 598: Tell us that this is a frozen class in the doc strings

src/votekit/pref_profile/pref_profile.py
```diff
@@ -653,59 +650,49 @@ class PreferenceProfile:
...
-        Also added protection for from_csv method.
+        else:
+            self.df, self.candidates_cast = self._init_from_df(df)
 
-        """
-        if self.candidates == tuple():
-            return self
+        self._find_max_ranking_length()
+        self._find_total_ballot_wt()
+        self._find_num_ballots()
```
peterrrock2, line 691: Have these return values and set them inside the init

src/votekit/pref_profile/pref_profile.py
```diff
@@ -891,6 +885,9 @@ def __init_formatted_df(
         if self.candidates != tuple() and contains_scores_indicator:
             col_order = list(self.candidates) + temp_col_order
         elif contains_scores_indicator:
+            for c in candidates_cast:
+                if c not in df.columns:
+                    df[c] = [np.nan] * len(df)
```
peterrrock2, line 890: Something like this. I don't remember if that syntax is valid. This just makes it clear what we are doing for the remaining candidates

```suggestion
                remaining_cands = set(candidates_cast) - set(df.columns)
                empty_df_cols = np.array(len(df), len(remaining_cands), fill=np.nan)
                df[list(remaining_cands)] = empty_df_cols
```

src/votekit/pref_profile/pref_profile.py
```diff
@@ -1000,30 +1092,80 @@ def find_max_ranking_length(self) -> Self:
...
+                b_row, self.candidates, self.max_ranking_length
+            )
+        return tuple(computed_ballots)
+
+    def __setattr__(self, name, value):
+        if getattr(self, "_is_frozen", False) and name != "_is_frozen":
+            raise AttributeError(
+                f"Cannot modify frozen instance: tried to set '{name}'"
+            )
+        super().__setattr__(name, value)
```
peterrrock2, line 1161: Let's not allow you to modify `_is_frozen` with a call like `prof._is_frozen`.

src/votekit/pref_profile/pref_profile.py
```diff
@@ -1053,6 +1195,21 @@ def group_ballots(self) -> PreferenceProfile:
...
             PreferenceProfile: A PreferenceProfile object with grouped ballot list.
         """
+        empty_df = pd.DataFrame(columns=["Voter Set", "Weight"], dtype=np.float64)
+        empty_df.index.name = "Ballot Index"
+
+        # if self.df.equals(empty_df):
+        #     return PreferenceProfile(
+        #         candidates=self.candidates,
+        #         max_ranking_length=self.max_ranking_length,
+        #     )
```
peterrrock2, line 1205:
```suggestion
```

peterrrock2 (changes requested):
Looking good so far!

peterrrock2 (changes requested):
Next step: change the nan to frozenset({'~'})

PR #234, opened 2025-07-10
Added basic STAR class with documentation

mvbbharath:
This PR adds a basic implementation of the STAR (Score Then Automatic Runoff) voting method to VoteKit. Only the "most top ratings" tiebreaker is implemented for now.


cjg266:
Hi @mvbbharath. I am going to put a PR with (1) test cases for your STAR and (2) a new STAR file with some edits. I had made a STAR, but I think yours is better overall fit with overall structure when compared to other elections. I did edit a few things, though or maybe I am just confused on some things), For example, I was wondering what is the best definition for remaining. As far as I understand, STAR only elects a single candidate, so I'm not sure it makes since that remaining are all the eliminated folks (unless we are also going to say the eliminated attribute for ElectionState state1 is the same as the remaining, which just seems slightly odd). Let me know what your think.

cdonnay:
Unless a candidate is actually eliminated by the election process (like in STV), we leave them as remaining. For example, I believe that in the `Plurality` election class all candidates except the winner are left as remaining. So it depends on how STAR treats them. Does the literature/algorithm outline say anything about it?

PR #242, opened 2025-07-11
3.3.0 bg rework

peterrrock2:


WillithG:
I believe this was a test PR as part of the "contributing to open source" tutorial, I.e. I think actual intended PRs should be seen in #245 and #246. Does that sound correct? If so, perhaps we can close this PR

PR #245, opened 2025-07-15
IC and IAC Ballot Generator improvements

WillithG:
This PR implements updated methods to sample profiles from IC and IAC. In particular, it de-couples the current behaviour of calling up to the parent `BallotSimplex` class and using the dirichlet distribution. Breaking changes were avoided by adding a flag


src/votekit/ballot_generator.py
```diff
@@ -430,6 +430,77 @@ class ImpartialCulture(BallotSimplex):
...
+            efficient way.
+
+            See BallotSimplex.generate_profile for method signature
+                description
+        '''
+        rng = np.random.default_rng()
+        ballots = [rng.permutation(self.candidates) for _ in range(number_of_ballots)]
+        return self.ballot_pool_to_profile(ballots, self.candidates)
+    
+    def generate_profile_MCMC(self, number_of_ballots: int, by_bloc: bool = False, BURN_IN_TIME = 0) -> PreferenceProfile | Dict:
```
WillithG, line 450: This method was more of an "exercise", it seems unlikely to be faster than `_generate_profile_optimized`, do we think it might be better to not add this method?

src/votekit/ballot_generator.py
```diff
@@ -430,6 +430,77 @@ class ImpartialCulture(BallotSimplex):
...
+        '''
+            Generate a preference profile for IC in a space and time
+            efficient way.
+
+            See BallotSimplex.generate_profile for method signature
+                description
+        '''
+        rng = np.random.default_rng()
+        ballots = [rng.permutation(self.candidates) for _ in range(number_of_ballots)]
+        return self.ballot_pool_to_profile(ballots, self.candidates)
```
WillithG, line 448: Example time data: with 25 cands and 10,000 ballots, %timeit gives the following
```
%timeit ic.generate_profile(number_of_ballots=10000, use_optimized=True)
1.54 s ± 106 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

src/votekit/ballot_generator.py
```diff
@@ -430,6 +430,85 @@ class ImpartialCulture(BallotSimplex):
...
+                adjacent-only swap of node
+            """
+            # the following line computes every possible
+            # neighbour-swap of given node
+            neighs = [
+                node[:i] + node[i + 1 : i - 1 : -1] + node[i + 2 :]
+                for i in range(1, len(node) - 1)
+            ]
+            neighs.append(node[1::-1] + node[2:])
+            return neighs
```
peterrrock2, line 484: Style: move this out and prefix with dunder `__compute_neighs`

peterrrock2: This needs to be extended to non-complete rankings as well: use integers to map to the correct type of operation

src/votekit/ballot_generator.py
```diff
@@ -430,6 +430,85 @@ class ImpartialCulture(BallotSimplex):
...
+        # initialize current ballot at some starting node
+        # for each i in {num of ballots}
+        # compute all n-1 swaps for current ballot
+        # uniformally choose one of the n-1 swaps to step to next
+        # record the destination of next step
+        num_cands = len(self.candidates)
+        ballot_ind = np.zeros((number_of_ballots, num_cands), dtype=np.int8)
+        # initialize starting node
+        next_node = list(range(num_cands))
+        random.shuffle(next_node)
```
peterrrock2, line 495: Add the IAC function in here that allows you to turn a integer into any possible ballot on n candidates

src/votekit/ballot_generator.py
```diff
@@ -430,6 +430,85 @@ class ImpartialCulture(BallotSimplex):
...
+
+        # burn in loop
+        for i in range(BURN_IN_TIME):
+            neighs = compute_neighs(next_node)
+            next_node = random.choice(neighs)
```
peterrrock2, line 500: Why do this if we are starting from a random ballot? We are sampling from a uniform distribution

peterrrock2 (commented):
This is a good first draft! We'll keep talking about this later

tests/ballot_generators/test_bg_distribution.py
```diff
@@ -133,6 +175,38 @@ def test_iac_distribution():
     assert do_ballot_probs_match_ballot_dist(ballot_prob_dict, generated_profile)
 
 
+# NOTE: Enable this test once the optimized version is completed
+def test_iac_optimized_distribution():
```
peterrrock2, line 179: You can use a pytest decorator to mark this as in progress (so you can skip it when running tests). That is better than just having it be true for a while because then we are less likely to forget to go back and fix the test.

peterrrock2 (commented):
Some more things for us to think about here!

src/votekit/ballot_generator.py
```diff
@@ -444,8 +553,65 @@ class ImpartialAnonymousCulture(BallotSimplex):
...
+            num_gaps, dtype=int
+        )  # record the number of gaps in stars/bars
+
+        # rather than iterate n! times, we perform multiple
+        # multinomial experiments
+        num_valid_ballots = self._total_ballots(num_cands, max_ballot_length)
+        num_iterations = num_valid_ballots // self._MAX_BINOM_EXPERIMENT_SIZE
+        remainder = num_valid_ballots % self._MAX_BINOM_EXPERIMENT_SIZE
+        pvals = [1/num_gaps] * num_gaps
+        for _ in range(num_iterations):
```
WillithG, line 605: Re: [Previous Thread](https://github.com/mggg/VoteKit/pull/245#discussion_r2211590724)
Here is an optimization suggested by @prismika, where instead of iterating $n!$ times and sampling a single index we iterate $n!/(2^63-1)$ times and draw from the multinomial repeatedly. 
I need to run some experiments to see when / if this runs better than the previous solution (since sampling random integers is fast).

In any case, we should probably have a discussion about this particular part of the process as @peterrrock2 suggested in the previous thread linked above

src/votekit/ballot_generator.py
```diff
@@ -430,6 +474,37 @@ class ImpartialCulture(BallotSimplex):
     def __init__(self, **data):
         super().__init__(alpha=float("inf"), **data)
 
+    def generate_profile(
+        self, number_of_ballots: int, by_bloc: bool = False, use_optimized=False,
+        max_ballot_length = None
```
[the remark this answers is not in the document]
WillithG: Update: I have updated the code to support "fixed length ballots only & specifying ballot length". I have done this with `np.random.choice()` (which chatgpt claims also uses Fisher-Yates), however, the exact algorithm used should be double-checked so that we can make sure this scales well

cdonnay:
Please change merge into wip/3.3.0-bg-module

WillithG:
Notes from discussion about caching `\tau_n` and `\mathcal B_n`
- change implementation to use class variables rather than instance variables, and then re-run the same timing experiments, and memory overhead tests
- no need for double caching --- we can use syntax like `b_n = _total_ballots(i) // (i)` (where `_total_ballots` is cached) and python will leverage the ballots cache for computing `\mathcal B_n`

WillithG:
I have changed the `index_to_ballot` caching behaviour so that: the cache is saved as a class variable and not an instance variable, and so that we are using a single cache instead of a double cache. Here are some preliminary run-time plots. I will check the memory overhead and push to higher number of ballots soon.

Note that in these runs, I skipped the step where `generate_profile` casts the samples to `Ballot` objects. I did this to save time in running these experiments, however, skipping this step should add the same amount of time to both the cached and uncached methods. 
`reset_cache=True` means that the cache was cleared between each sampled point. 

image

image

WillithG:
Notes from offline discussion:
- perform same timing experiments against "bare" `_index_to_ballot` to acquire lower bound for IC runtimes
- optimize storage and return of ballot samples by instantiating returned `PreferenceProfile` via df

WillithG:
Most recent changes implement some of the refactoring that we've discussed for IC and IAC. In particular,
- The IC and IAC classes no longer inherit from `BallotSimplex`, and the option to sample using Dirichlet is no longer available
- We now generate resultant `PreferenceProfiles` by building a df, rather than converting the sampled Ballots to `Ballot` objects
- The methods shared across them have now been moved to `utils.py`
  - (namely, Peter's `index_to_lexicographic_ballot` and a method for converting sampled ballot indices to a valid df for Preference profiles
- I have not yet moved the IC and IAC classes into their own file, as I imagine there should be a brief discussion about what the new directory structure should be (e.g. should we move each of these classes to their own file under a new directory `/src/ballot_genertors/`, or would we like another solution?)

peterrrock2:
Thank you for your work on this! We'll keep editing this as we move forward with our FE stuff

Issue #236, opened 2025-07-10
Explore and Update MCMC for BT


WillithG:
The following code block and image shows that the MCMC method for generating BT ballots does not always deliver the same distribution of ballots compared to computing the pdf directly. This issue involves exploring parameter spaces of bad mixing and implementing upgrades (such as burn in time, Metropolis-Hastings on the shortcut ballot graph) to help with this issue

(e.g. 10,000 ballots, four candidates, two slates, two blocs)
```
cands = ["W1", "W2", "C1", "C2"]
bt = name_BradleyTerry(
    candidates=cands,
        pref_intervals_by_bloc={
            "W": {
                "W": PreferenceInterval({"W1": 0.6, "W2": 0.3}),
                "C": PreferenceInterval({"C1": 0.05, "C2": 0.05}),
            },
            "C": {
                "W": PreferenceInterval({"W1": 0.2, "W2": 0.2}),
                "C": PreferenceInterval({"C1": 0.3, "C2": 0.3}),
            },
        },
        bloc_voter_prop={"W": 0.5, "C": 0.5 },
        cohesion_parameters={"W": {"W": 0.5, "C": 0.5}, "C": {"C": 0.9, "W": 0.1}},
    )

num_ballots = 10000
bals_BT_classic = bt.generate_profile(num_ballots)
bals_BT_MCMC = bt.generate_profile_MCMC(num_ballots, verbose=False)

...
```

(Blue bars are ballots for which MCMC oversampled, beige bars are ballots that MCMC undersampled. Brown bars are ballots for where MCMC and direct computation of the pdf agree)

<img width="684" height="475" alt="Image" src="https://github.com/user-attachments/assets/c2bd6401-ac0b-4094-97aa-924b11690900" />

WillithG:
Working branch [](https://github.com/mggg/VoteKit/tree/3.3.0-bg-rework)

ppeaung:
Also working on studying the mixing time for the burn in time for MCMC. Here are some preliminary experiments on the total variation with the calculated pdf (ideally near zero). 

<img width="576" height="455" alt="Image" src="https://github.com/user-attachments/assets/987b0812-7ab3-42cc-a29a-2cc4c87ea377" />

Implementing a subroutine of what burn-in time would be appropriate with the number of candidates would be nice as well.

PR #246, opened 2025-07-21
Name Bradley-Terry Sub-sampling

ppeaung:
This PR implements sub-sampling for the Metropolis-Hastings MCMC algorithm for name Bradley-Terry ballots. In particular, it takes in `chain_length` and `number_of_ballots` as parameters. `chain_length` is the length of the Markov Chain which we sample from and `number_of_ballots` is the number of samples, spaced evenly. For example, if we have `chain_length = 100` and `number_of_ballots = 10`, we draw every other 10 ballots in the chain. Some unit tests have also been implemented in as well. This partially addresses issue #236.


src/votekit/ballot_generator.py
```diff
@@ -834,9 +965,113 @@ def generate_profile(
         # else return the combined profiles
         else:
             return pp
+
```
ppeaung, line 968: Helper function for the random walk on the short cut graph; instead of swapping neighboring rankings, we swap any indices.

src/votekit/ballot_generator.py
```diff
@@ -834,9 +965,113 @@ def generate_profile(
...
+            raise ValueError("Some element of ballots list is not a ballot.")
+
+        pp = PreferenceProfile(ballots=ballots)
+        pp = pp.group_ballots()
+        return pp
+
 
     def _BT_mcmc(
-        self, num_ballots, pref_interval, seed_ballot, zero_cands={}, verbose=False
+        self, num_ballots, pref_interval, seed_ballot, zero_cands={}, verbose=False, BURN_IN_TIME=0
```
ppeaung, line 1074: We added in a defaulted burn-in time which explores the state space and allows the chain to forget its starting point. This can either be implemented or dropped.

peterrrock2:
Thanks for your work on this! We'll go ahead and merge this in and keep working on it as we do the refactor of the ballot generators

PR #276, opened 2025-08-14
3.3.0 metric distances update

peterrrock2:
Change the way that we compute EMD for preference profiles. We now avoid using floyd-warshall, the BallotGraph class, and the OT library. Instead, we solve EMD via a scipy linear program, and this allows us to compute the EMD between profiles with many more candidates.

### DIFF
--- a/tests/metrics/test_emd.py
+++ b/tests/metrics/test_emd.py
@@ -271,7 +271,7 @@ def test_new_emd_speed():
         total_time += time() - start_time
 
     # This is a super wide bound. On my machine, this takes 1.2 seconds, but hardware can really
-    # change this, and < 1s is totally fine since this used to take over an hour.
+    # change this, and < 10s is totally fine since this used to take over an hour.
     assert total_time < 10
 
 


peterrrock2:
I rebased before I opened this, and it looks like there are some mypy and test errors from another PR that didn't get resolved. Let me know if you want me to fix them here or if you want another PR with the fixes

src/votekit/metrics/distances.py
```diff
@@ -1,45 +1,384 @@
...
+            indices representing rankings and values are tuples of weights from each profile
+            in the order of pp1 and pp2.
     """
-    # create ballot graph
-    ballot_graph = BallotGraph(source=pp2).graph
-    # ballot_graph = graph.from_profile(profile=pp2, complete=True)
+    profile1 = pp1.group_ballots()
+    profile2 = pp2.group_ballots()
+
+    cand_to_index_mapping = {
```
cdonnay, line 283: Do you need to check that both profiles have the same candidates?

src/votekit/metrics/distances.py
```diff
@@ -1,45 +1,384 @@
...
-    electB_distr = np.array(em_array(pp=pp2))
+    for idx, ranking_tuple in enumerate(profile1_ranking_array):
+        tup = tuple(
+            [
+                cand_to_index_mapping[cand]
+                for cand_set in ranking_tuple
+                for cand in cand_set
+                if cand != "~"
+            ]
+        )
```
cdonnay, line 304: I think that you need to do some kind of check for empty frozensets, which would not appear in the list of candidates, but denote skipped positions in the ranking. 

It's probably the case that we can't compute the EMD for profiles which contain ballots with skips. I don't know how those ballots would be represented on the ballot graph.

src/votekit/metrics/distances.py
```diff
@@ -1,45 +1,384 @@
...
+
+    for idx, ranking_tuple in enumerate(profile2_ranking_array):
+        tup = tuple(
+            [
+                cand_to_index_mapping[cand]
+                for cand_set in ranking_tuple
+                for cand in cand_set
+                if cand != "~"
+            ]
+        )
```
cdonnay, line 322: Same concern here; empty frozensets

src/votekit/metrics/distances.py
cdonnay: Ah I see you validate that before passing here, and I guess it actually is a totally valid computation even with disjoint candidate sets.

src/votekit/metrics/distances.py
```diff
@@ -1,45 +1,384 @@
...
+        where I_n is the identity matrix of size n (number of sources) and I_m is the identity
+        matrix of size m (number of targets). The linear program can be expressed as:
+
+        minimize c^T x
+
+        subject to:
+            Ax = y
+
+        where 'y' is the concatenation of the source and target distributions y = [a, b]^T.
+
```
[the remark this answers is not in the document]
peterrrock2: I am using trying to use the quotes when math is internal to a sentence and not when it is on its own line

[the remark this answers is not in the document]
peterrrock2: I see what you mean, though. I'll go with quotes on all things inside of sentences

src/votekit/metrics/distances.py
peterrrock2: You got it on the money. This function is entirely internal too, so the validation is skipped since I do it before getting here

src/votekit/metrics/distances.py
peterrrock2: Good point. Thank you!

tests/metrics/test_emd.py
```diff
@@ -247,3 +232,79 @@ def test_emd_profile_errors():
...
+    for _ in range(10):
+        profile1 = make_random_profile(n_voters, alphabet)
+        profile2 = make_random_profile(n_voters, alphabet)
+
+        start_time = time()
+        _ = earth_mover_dist(profile1, profile2)
+        total_time += time() - start_time
+
+    # This is a super wide bound. On my machine, this takes 1.2 seconds, but hardware can really
+    # change this, and < 1s is totally fine since this used to take over an hour.
```
cdonnay, line 274:
```suggestion
    # change this, and < 10s is totally fine since this used to take over an hour.
```

cdonnay (approved):

PR #278, opened 2025-08-15
Fix #269

### LINKED ISSUE #269: Memory Leak in RankedPairs
I did this. I'm fixing it. TL;DR, the "store_state" flag matters when you run `get_profile`. This shouldn't matter, but is a package-wide thing to fix in future release

peterrrock2:
Turns out "store_states" matters only when you run "get_step"


cdonnay (approved):
Nice!

PR #279, opened 2025-08-15
Bugfix/mypyerros 3.3.0

peterrrock2:
Fix some mypy errors and mark some tests as slow while I'm at it

### DIFF
--- a/src/votekit/plots/bar_plot.py
+++ b/src/votekit/plots/bar_plot.py
@@ -103,13 +103,13 @@ def _set_default_bar_plot_args(
     if isinstance(threshold_values, list):
         if not isinstance(threshold_kwds, list):
             threshold_kwds = [dict() for _ in range(len(threshold_values))]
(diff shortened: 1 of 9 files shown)


src/votekit/plots/bar_plot.py
```diff
@@ -284,15 +284,23 @@ def _prepare_data_bar_plot(
...
-    all_data_values = []
+    all_data_values: list[float] = []
     # NOTE: data is dict[str, dict[str, float]]
     for dct in data.values():
         if isinstance(dct, dict):
             all_data_values.extend(dct.values())
 
     flat_y_data = [item for sublist in y_data for item in sublist]
+    print(flat_y_data)
+    print(all_data_values)
```
cdonnay, line 295:
```suggestion
```

cdonnay (changes requested):
One quick change then you're good!

Issue #232, opened 2025-07-02
STV with numpy


cdonnay:
Here is an idea for a faster implementation of ranked elections with no tied ballots; specifically STV. 

The thing that I would do if I trying to make `STV` with `numpy` (assuming no tied positions) would be to start with some array of int8 types (less memory) where each positive value is mapped to a candidate and then the value -127 would be mapped to "no candidate". Then, I would create 4 data structures, a fpv vector, a weight vector, a position vector, and a ballot matrix of dimension (n_distinct_ballots)x(max_ballot_length+1). You can then get the weight for a fpv using `np.bincount(fpv_vec, wieghts=wt_vec)` so you can quickly find the person with the most and the least votes. We then go to elect or eliminate a candidate. Use `elected_or_eliminated_indexes = np.where(fpv_vec==elected_or_eliminated_id)` to find the indexes where that candidate was the fpv and then take `pos_vector[elected_or_eliminated_indexes] += 1` . Then you can set `fpv_vec = ballot_mat[:, pos_vector]` since that will grab the appropriate rows. Now update the `wt_vec` using the the `elected_or_eliminated_indexes` according to your transfer rule.

PR #286, opened 2025-08-25
Fast STV

EdouardHeitzmann:
This is an alternative implementation of the STV algorithm using numpy as suggested in #232.

This is a complete bottom-up re-work of the STV class. Because the numpy machinery at the core of this implementation is incompatible with the `Election` and `RankingElection` structure which builds a new `PreferenceProfile` for each round, this implementation builds a new class from scratch. I did however give this new class all the same methods of the old `STV` class, and its output is formatted in exactly the same way.

This implementation is around ~30x faster than the old one for large profiles; the main caveat being that the `get_profile` method is slower for the early rounds of large elections, since the new implementation builds the profile from scratch every time it is requested, even when this profile is very similar to the one it started with. Apart from building profiles, the numpy implementation is faster.

Because the machinery behind this implementation no longer iterates over a list of `Ballot` objects in every round of an election, it is no longer able to easily accept a callable `transfer` argument. Instead, it accepts a string `transfer` argument, with the only current options being 'fractional' and 'random.' The 'random' option is also slightly different from the old `random_transfer` callable from votekit, which would break if there were fewer transferable ballots than surplus ballots in a round.

The output formatting and round-by-round scoring of this implementation has been tested extensively; the round-by-round tallies agree with the old STV implementation to within a precision of 10**(-5) in all tests so far (which is to say: the test suite, the Portland elections, and the Scottish elections). Some more testing might be required for these features, however:
- Check that the borda tiebreak option is working correctly;
- Check that the random transfer option is working correctly.

Note that this is not the fastest this implementation could get -- especially if the goal is to run many elections iteratively where only the `get_elected` or `get_ranking` methods are needed, there's still a lot of fat that could be trimmed away. In particular, it might be nice to separate the `_run_STV` method driving this implementation as a separate function outside of this class, in case users want to avoid converting back and forth with a numpy matrix every time they iterate an election. Potential thoughts for where to go next!


src/votekit/elections/election_types/ranking/stv.py
```diff
@@ -16,6 +17,705 @@
...
+            ]
+        )
+
+    def get_status_df(self, round_number: int = -1) -> pd.DataFrame:
+        """
+        Yield the status (elected, eliminated, remaining) of the candidates in the given round.
+        DataFrame is sorted by current ranking.
+
+        Args:
+            round_number (int, optional): The round number. Supports negative indexing. Defaults to
```
cdonnay, line 467:
```suggestion
            round_number (int): The round number. Supports negative indexing. Defaults to
```

src/votekit/elections/election_types/ranking/stv.py
```diff
@@ -16,6 +17,705 @@
...
+                status_df.at[self.candidates[birthday_list[0]], "Status"] = "Eliminated"
+                status_df.at[self.candidates[birthday_list[0]], "Round"] = turn_id + 1
+            elif turn_type == 1:  # winner
+                for c in birthday_list:
+                    status_df.at[self.candidates[c], "Status"] = "Elected"
+                    status_df.at[self.candidates[c], "Round"] = turn_id + 1
+            elif turn_type == 2:  # winner by default
+                for c in birthday_list:
+                    status_df.at[self.candidates[c], "Status"] = "Elected"
+                    status_df.at[self.candidates[c], "Round"] = turn_id + 1
```
cdonnay, line 502: Super unclear to me what's happening here. Descriptive variable names please.

src/votekit/elections/election_types/ranking/stv.py
```diff
@@ -16,6 +17,705 @@
...
+                            frozenset([self.candidates[c] for c in play[1]])
+                            for c in play[1]
+                        ),
+                        eliminated=(frozenset(),),
+                        scores={
+                            self.candidates[c]: self._tally_record[i + 1][c]
+                            for c in self._tally_record[i + 1].nonzero()[0]
+                        },
+                    )
+                )
```
cdonnay, line 625: Same comment as _run_STV; modularize this.

It's also a red flag to me when I see things like `if play[-1] == 2`, just because that is super opaque.

src/votekit/elections/election_types/ranking/stv.py
```diff
@@ -16,6 +17,705 @@
...
+
+        return e_states
+
+    def get_profile(self, round_number: int = -1) -> PreferenceProfile:
+        """
+        Fetch the PreferenceProfile of the given round number.
+
+        Args:
+            round_number (int, optional): The round number. Supports negative indexing. Defaults to
```
cdonnay, line 634:
```suggestion
            round_number (int): The round number. Supports negative indexing. Defaults to
```

src/votekit/elections/election_types/ranking/stv.py
```diff
@@ -16,6 +17,705 @@
...
+        for i, row in enumerate(self._ballot_matrix):
+            ballot = []
+            for entry in row:
+                if entry == -127:
+                    break
+                elif entry in hopeful:
+                    ballot.append(frozenset([self.candidates[entry]]))
+            if len(ballot) > 0:
+                ballots.append(Ballot(ranking=tuple(ballot), weight=wt_vec[i]))
+        return condense_profile(
```
cdonnay, line 668: Why do you have to call condense profile? Are you adding in skipped rankings?

src/votekit/elections/election_types/ranking/stv.py
```diff
@@ -16,6 +17,705 @@
...
+            PreferenceProfile(
```
cdonnay, line 669: Can you construct the profile using the df init method instead? rather than creating a list of ballots?

src/votekit/elections/election_types/ranking/stv.py
```diff
@@ -16,6 +17,705 @@
...
+        hopeful = self._tally_record[round_number].nonzero()[0]
+        ballots = []
+        wt_vec = self._wt_vec.copy()
+        if self.m > 1:
+            # find the last entry in play_by_play[:round_number] with a 1 in the last position (there may be no such entry)
+            for i in range(len(self._play_by_play[:round_number]) - 1, -1, -1):
+                if self._play_by_play[i][-1] == 1:
+                    wt_vec = self._play_by_play[i][2]
+                    break
+        # print(wt_vec)
```
cdonnay, line 658:
```suggestion
```

src/votekit/elections/election_types/ranking/stv.py
```diff
@@ -16,6 +17,705 @@
...
+        """
+        if (
+            round_number < -len(self.election_states)
+            or round_number > len(self.election_states) - 1
+        ):
+            raise IndexError("round_number out of range.")
+
+        round_number = round_number % len(self.election_states)
+
+        hopeful = self._tally_record[round_number].nonzero()[0]
```
cdonnay, line 649: Unclear variable name.

src/votekit/elections/election_types/ranking/stv.py
```diff
@@ -16,6 +17,705 @@
...
+        ballots = []
+        wt_vec = self._wt_vec.copy()
+        if self.m > 1:
+            # find the last entry in play_by_play[:round_number] with a 1 in the last position (there may be no such entry)
+            for i in range(len(self._play_by_play[:round_number]) - 1, -1, -1):
+                if self._play_by_play[i][-1] == 1:
+                    wt_vec = self._play_by_play[i][2]
```
cdonnay, line 656: What's happening here? These integer valued things are really hard to read.

src/votekit/elections/election_types/ranking/stv.py
```diff
@@ -16,6 +17,705 @@
...
+        )
+
+    def get_step(
+        self, round_number: int = -1
+    ) -> tuple[PreferenceProfile, ElectionState]:
+        """
+        Fetches the profile and ElectionState of the given round number.
+
+        Args:
+            round_number (int, optional): The round number. Supports negative indexing. Defaults to
```
cdonnay, line 681:
```suggestion
            round_number (int): The round number. Supports negative indexing. Defaults to
```

src/votekit/elections/election_types/ranking/stv.py
```diff
@@ -16,6 +17,705 @@
...
+                -1, which accesses the final profile.
+
+        Returns:
+            tuple[PreferenceProfile, ElectionState]
+        """
+        return (self.get_profile(round_number), self.election_states[round_number])
+
+    def _sample_to_transfer(
+        self, fpv_vec, wt_vec, w: np.int8, s: int, rng=None
```
cdonnay, line 690: type setting

src/votekit/elections/election_types/ranking/stv.py
```diff
@@ -16,6 +17,705 @@
...
+    ) -> np.ndarray:
+        """
+        Build a list of indices i such that fpv_vec[i] == w;
+        each index is repeated round(wt_vec[i]) times, then we sample s entries uniformly.
```
cdonnay, line 694: Doc strings

src/votekit/elections/election_types/ranking/stv.py
```diff
@@ -16,6 +17,705 @@
...
+        This is different from the Cambridge transfer!!!
+        Some of the ballots selected for transfer may still be exhausted if they list no further preference.
```
cdonnay, line 696: We definitely want to implement the Cambridge version. To my understanding, in that version, if a ballot is selected to be transferred but is exhausted, then they select another ballot. It is true that if a lot of ballots are bullet/short, this could mean that not enough ballots exist to transfer, but in that case we should raise a warning saying what happened. Cambridge does not describe what to do in that edge case; it probably has never happened in a real, large election.

https://opavote.com/methods/cambridge-stv-rules

cdonnay (changes requested):
Hi Ed! Great start here. Have a bunch of tiny doc string comments, along with some bigger organizational pieces. The TLDR is we want to modularize some of these long functions to make them easier to read. In general, lots of your variables have opaque names/don't make use of Python structures that would make them easier to follow.

Also be sure to run `poetry run mypy src`, I found 17 errors in this file that you need to address.

src/votekit/elections/election_types/ranking/stv.py
```diff
@@ -16,6 +17,705 @@
...
+        self._winners, self._tally_record, self._play_by_play, self._tiebreak_record = (
+            self._run_STV(
+                self._ballot_matrix,
+                np.array(self._wt_vec),
+                self._fpv_vec,
+                m,
+                len(self.candidates),
+            )
+        )
+        self.election_states = self._election_states()
```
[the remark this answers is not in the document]
EdouardHeitzmann: It is creating them -- so maybe _make_election_states()?

src/votekit/elections/election_types/ranking/stv.py
```diff
@@ -16,6 +17,705 @@
...
+
+    def _stv_validate_profile(
+        self, profile: PreferenceProfile
+    ):  # this might be adding overhead in general (~100ms)? Maybe add a validate_profile = False param in init
```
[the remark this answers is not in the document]
EdouardHeitzmann: Cool! I was worried about this because I was in the mindframe of running many thousands of these STV elections in an MCMC chain; but I can resolve those concerns somewhere else.

src/votekit/elections/election_types/ranking/stv.py
```diff
@@ -16,6 +17,705 @@
...
+        quota: str = "droop",
+        simultaneous: bool = False,
+        tiebreak: Optional[str] = None,
+    ):
+        self._stv_validate_profile(profile)
+
+        if m <= 0:
+            raise ValueError("m must be positive.")
+        elif len(profile.candidates_cast) < m:
+            raise ValueError("Not enough candidates received votes to be elected.")
```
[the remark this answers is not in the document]
EdouardHeitzmann: Will do!

src/votekit/elections/election_types/ranking/stv.py
```diff
@@ -16,6 +17,705 @@
...
+                                fpv_vec[i] = ballot_matrix[i, pos_vec[i]]
+                                wt_vec[i] = new_weights[i]
+                    play_by_play.append((turn, [w], np.array(wt_vec), 1))
+                    turn += 1
+                tallies = np.bincount(
+                    fpv_vec[fpv_vec != -127],
+                    weights=wt_vec[fpv_vec != -127],
+                    minlength=ncands,
+                )
+                tally_record.append(tallies.copy())
```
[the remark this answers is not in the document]
EdouardHeitzmann: Cool! this makes a lot of sense in hindsight -- I honestly just haven't had a real look at the main engine as a whole in several weeks, and some of the extra nesting happened as I was adding extra features to deal with tiebreaks etc. I have a pretty good sense of how to break this code up!

src/votekit/elections/election_types/ranking/stv.py
```diff
@@ -16,6 +17,705 @@
...
+                L = np.argmin(masked_tallies)
+            gone_list.append(L)
+            for i in range(len(fpv_vec)):
+                if fpv_vec[i] == L:
+                    while ballot_matrix[i, pos_vec[i]] in gone_list:
+                        pos_vec[i] += 1
+                    fpv_vec[i] = ballot_matrix[i, pos_vec[i]]
+            play_by_play.append((turn, [L], [], 0))
+            turn += 1
+        return winner_list, tally_record, play_by_play, tiebreak_record
```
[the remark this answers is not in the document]
EdouardHeitzmann: At a high level this block of the code tries to eliminate a loser if the election is not over or if nobody has quota. The multiple return statements happen because there are multiple ways the election can finish -- although now that I'm looking at it I can probably remove the very bottom one, since the election never ends with an elimination.

I'll get to work on helper functions!

src/votekit/elections/election_types/ranking/stv.py
```diff
@@ -16,6 +17,705 @@
...
+        Args:
+            round_number (int, optional): The round number. Supports negative indexing. Defaults to
+                -1, which accesses the final profile.
+
+        Returns:
+            tuple[frozenset[str],...]:
+                Tuple of sets of remaining candidates. Ordering of tuple
+                denotes ranking of remaining candidates, sets denote ties.
+        """
+        tallies = self._tally_record[round_number]
```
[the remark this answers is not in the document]
EdouardHeitzmann: Oh weird I thought everybody used this word -- yeah a tally is just something that's been added up. Usually by drawing little groups of five sticks on a piece of paper?

src/votekit/elections/election_types/ranking/stv.py
EdouardHeitzmann: Cool! this sounds smart.

src/votekit/elections/election_types/ranking/stv.py
EdouardHeitzmann: The short answer is I needed to do that to get matching output to the old STV function. 

The longer answer is, the way I iterate through my ballot matrix, multiple initially different ballots might become the same one, after some candidates are removed from it, so I would end up with lots of different ballots with different weights in my profile, which is why my output was not matching the old STV's profiles.

src/votekit/elections/election_types/ranking/stv.py
```diff
@@ -16,6 +17,705 @@
...
+        self.threshold = self._get_threshold(profile.total_ballot_wt)
+        self.simultaneous = simultaneous
+        self.tiebreak = tiebreak
+
+        self.candidates = list(
+            profile.candidates
+        )  # canonical ordering! don't ever touch this list again
+
+        self._ballot_matrix, self._wt_vec, self._fpv_vec = self._convert_df(profile)
+        self._winners, self._tally_record, self._play_by_play, self._tiebreak_record = (
```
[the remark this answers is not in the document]
cdonnay: Maybe `fpv_by_round`?

src/votekit/elections/election_types/ranking/stv.py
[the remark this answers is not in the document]
cdonnay: Great!

src/votekit/elections/election_types/ranking/stv.py
```diff
@@ -188,114 +178,104 @@ def _run_STV(
...
+        
+        #below is the main loop of the algorithm
         while len(winner_list) < m:
             # force the bincount to count entries 0 through ncands-1, even if some candidates have no votes
-            tallies = np.bincount(
-                fpv_vec[fpv_vec != -127],
-                weights=wt_vec[fpv_vec != -127],
-                minlength=ncands,
-            )  # don't count -127 entries from fpv_vec
+            tallies = make_tallies(fpv_vec, wt_vec, ncands)
```
cdonnay, line 271: Does it make sense to call this something involving "fpv"?

tests/elections/election_types/ranking/test_fastSTV_extended.py
```diff
@@ -0,0 +1,112 @@
...
+def is_same_election_state(e1, e2):
+    return (
+        e1.round_number == e2.round_number
+        and e1.remaining == e2.remaining
+        and e1.elected == e2.elected
+        and e1.eliminated == e2.eliminated
+        and e1.tiebreaks == e2.tiebreaks
+        and e1.scores.keys() == e2.scores.keys()
+        and all(np.isclose(e1.scores[k], e2.scores[k]) for k in e1.scores.keys())
+    )
```
cdonnay, line 18: I think the pydantic dataclass automatically has __eq__ built in.

tests/elections/election_types/ranking/test_fastSTV_extended.py
```diff
@@ -0,0 +1,112 @@
...
+
+def check_full_election(profile):
+    new = fast_STV(profile, m=2, simultaneous=simult)
+    old = STV(profile, m=2, simultaneous=simult)
+
+    num_rounds = len(new._tally_record)
+    
+    for i in range(num_rounds):
+        # check that get_rankings returns the same pandas df string
+        if old.get_status_df(i).to_string(index=True, justify="justify") != new.get_status_df(i).to_string(index=True, justify="justify"):
```
cdonnay, line 51: There should be a pandas df eq method

tests/elections/election_types/ranking/test_fastSTV_extended.py
```diff
@@ -0,0 +1,112 @@
...
+            print("Old:")
+            print(old.get_ranking(i))
+            print("New:")
+            print(new.get_ranking(i))
+            print("Round number:", i)
+            return True
+
+def test_scot_methods():
+    for filename in tqdm(os.listdir("data/8_cands")):
+        if filename.endswith(".csv"):
```
cdonnay, line 98: I think you can do something like `"data/8_cands/*.csv"` to remove this if statement.

tests/elections/election_types/ranking/test_fastSTV_extended.py
```diff
@@ -0,0 +1,112 @@
...
+            pf1 = load_scottish(f"data/8_cands/{filename}")[0]
+            if check_full_election(pf1):
+                print(f"Election {filename} failed")
+                assert False
+    assert True
```
cdonnay, line 103: Redo the T/F so that you can just say `assert check_full_election(pf1)`.

tests/elections/election_types/ranking/test_fastSTV_extended.py
```diff
@@ -0,0 +1,112 @@
...
+
+def test_scot_states():
+    for filename in tqdm(os.listdir("data/8_cands")):
+        if filename.endswith(".csv"):
+            pf1 = load_scottish(f"data/8_cands/{filename}")[0]
+            if check_election_states(pf1):
+                print(f"Election {filename} failed")
+                assert False
+    assert True
```
cdonnay, line 112: Same here, change T/F so that you can just assert check_election_states

tests/elections/election_types/ranking/test_fastSTV_extended.py
```diff
@@ -0,0 +1,112 @@
...
+            print("Round number:", i)
+            return True
+        # check that get_remaining returns the same thing
+        if old.get_remaining(i) != new.get_remaining(i):
+            print("Mismatch found during get_remaining!")
+            print("Old:")
+            print(old.get_remaining(i))
+            print("New:")
+            print(new.get_remaining(i))
+            print("Round number:", i)
```
cdonnay, line 75: print statements might have been helpful for debuggin but I think that pytests traditionally do not have print statements... is that true @peterrrock2 ?

i.e. you don't need to have all these if statements, just check if the df are equal and stop

tests/elections/election_types/ranking/test_fastSTV_extended.py
```diff
@@ -0,0 +1,112 @@
...
+    for filename in tqdm(os.listdir("data/8_cands")):
+        if filename.endswith(".csv"):
+            pf1 = load_scottish(f"data/8_cands/{filename}")[0]
+            if check_full_election(pf1):
+                print(f"Election {filename} failed")
+                assert False
+    assert True
+
+def test_scot_states():
+    for filename in tqdm(os.listdir("data/8_cands")):
```
cdonnay, line 106:
```suggestion
    for filename in tqdm(os.listdir("data/8_cands/*.csv")):
```

tests/elections/election_types/ranking/test_fastSTV_extended.py
```diff
@@ -0,0 +1,112 @@
...
+
+    for i, (ns, os) in enumerate(zip(new_states, old_states)):
+        if not is_same_election_state(ns, os):
+            print("Mismatch found during election_states!")
+            print("Old:")
+            print(os)
+            print("New:")
+            print(ns)
+            print("State number:", i)
+            return True
```
cdonnay, line 41: If i'm right about pydantic eq methods, then all this is `if new_states == old_states`.

src/votekit/elections/election_types/ranking/stv.py
```diff
@@ -188,114 +178,104 @@ def _run_STV(
...
-                tallies = np.bincount(
-                    fpv_vec[fpv_vec != -127],
-                    weights=wt_vec[fpv_vec != -127],
-                    minlength=ncands,
-                )
+                winners = find_winners(self)
+                update_because_winner(self)
+                play_by_play.append((turn, winners, np.array(wt_vec), 1))
+                turn += 1
+                tallies = make_tallies(fpv_vec, wt_vec, ncands)
```
cdonnay, line 278: Am I correct in seeing that this only does simultaneous, never 1by1? or the other way around?

cdonnay (changes requested):
Let's replace the closures inside STV with private methods instead!

src/votekit/elections/election_types/ranking/stv.py
```diff
@@ -16,6 +17,705 @@
...
+                        -127
+                    )  # possible TODO: improve this iteration not to look at ballot entries beyond the first empty one -- maybe by removing rows from the df as we go?
+                else:
+                    try:
+                        ballot_matrix[row_idx, col_idx] = candidate_to_index[val]
+                    except KeyError:
+                        raise ValueError(f"Candidate '{val}' not in candidate list.")
+
+        wt_vec[:] = df["Weight"].astype(np.float64).to_numpy()
+        fpv_vec = ballot_matrix[:, 0].copy()
```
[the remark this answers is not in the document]
EdouardHeitzmann: (I think)

src/votekit/elections/election_types/ranking/stv.py
```diff
@@ -16,6 +17,705 @@
...
+                ).strip()  # unwrap the frozenset; I hecking love frozensets
+                if val == "~":
+                    ballot_matrix[row_idx, col_idx] = (
+                        -127
+                    )  # possible TODO: improve this iteration not to look at ballot entries beyond the first empty one -- maybe by removing rows from the df as we go?
+                else:
+                    try:
+                        ballot_matrix[row_idx, col_idx] = candidate_to_index[val]
+                    except KeyError:
+                        raise ValueError(f"Candidate '{val}' not in candidate list.")
```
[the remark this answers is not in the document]
EdouardHeitzmann: You were right! cf. new _convert_df method.

src/votekit/elections/election_types/ranking/stv.py
```diff
@@ -16,6 +17,705 @@
...
+        """
+        tallies = self._tally_record[round_number]
+        # weird dict structure to detect ties, which must be put in the same fset
+        this_is_great_and_not_weird = dict()
+        for c, t in enumerate(tallies):
+            if t > 0:
+                if t not in this_is_great_and_not_weird:
+                    this_is_great_and_not_weird[t] = [self.candidates[c]]
+                else:
+                    this_is_great_and_not_weird[t].append(self.candidates[c])
```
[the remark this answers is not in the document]
EdouardHeitzmann: I don't see a way to do that (and chat didn't either)

src/votekit/elections/election_types/ranking/stv.py
```diff
@@ -16,6 +17,705 @@
...
+                they are tied.
+        """
+        if (
+            round_number < -len(self._tally_record)
+            or round_number > len(self._tally_record) - 1
+        ):
+            raise IndexError("round_number out of range.")
+        round_number = round_number % len(self._tally_record)
+        list_of_winners = [
+            [c]
```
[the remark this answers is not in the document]
EdouardHeitzmann: I don't think so!

src/votekit/elections/election_types/ranking/stv.py
EdouardHeitzmann: It did both! But it's gone now, it can't hurt us

tests/elections/election_types/ranking/test_fastSTV_extended.py
EdouardHeitzmann: I'm ~45% sure I still need this because the tallies are not always strictly equal

src/votekit/elections/election_types/ranking/stv.py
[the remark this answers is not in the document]
EdouardHeitzmann: I'm kind attached to this variable name actually :( it's what all the auditing literature uses

I'll change it if you insist though!

peterrrock2:
Converting to draft until we are ready to do the full pull (save resources on Test workflow)

src/votekit/elections/election_types/ranking/stv.py
```diff
@@ -7,15 +7,979 @@
...
+        df.insert(0, "Ballot Index", np.arange(n_rows, dtype=int))
+        df.set_index("Ballot Index", inplace=True)
+
+        # --- 7) Voter Set: empty set per row (distinct objects) ---
+        df["Voter Set"] = [set() for _ in range(n_rows)]
+
+        # --- 8) Weight column ---
+        df["Weight"] = wt_vec.astype(np.float64, copy=False)
+
+        return condense_profile(PreferenceProfile(contains_rankings=True, max_ranking_length=self.profile.max_ranking_length, candidates=tuple([self.candidates[c] for c in remaining]), df=df))
```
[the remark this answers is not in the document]
EdouardHeitzmann: The short answer is to pass the test suite! I think that without condensing we get multiple of the same type of ballot show up in the profile.

tests/elections/election_types/ranking/test_fast_STV.py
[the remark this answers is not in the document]
EdouardHeitzmann: there's none! I just renamed fast_STV to fastSTV recently.

EdouardHeitzmann:
All set! Typing and docstrings should be good now. You might decide to remove more comments, if you like.

src/votekit/elections/election_types/ranking/stv.py
```diff
@@ -16,6 +17,1011 @@
...
+
+        # this tells us who the new fpv vote will be with the updated positions
+        next_fpv_vec = self._ballot_matrix[winner_row_indices, next_fpv_pos_vec]
+        # don't update fpv_vec yet because we need to know current fpv to transfer weights
+
+        if self.transfer == "fractional":
+            get_transfer_value = np.frompyfunc(
+                lambda w: (tallies[w] - self.threshold) / tallies[w], 1, 1
+            )
+            transfer_value_values = get_transfer_value(
```
cdonnay, line 289: Name of variable and function are confusing

src/votekit/elections/election_types/ranking/stv.py
```diff
@@ -16,6 +17,1011 @@
...
+            tallies (np.ndarray): Current tallies for each candidate.
+            initial_tally (np.ndarray): Initial tallies for tiebreaking.
+            turn (int): The current round number.
+            mutant_bool_ballot_matrix (np.ndarray): Boolean mask for eliminated candidates.
+            mutant_winner_list (list[int]): List of winner candidate indices so far.
+            mutant_eliminated_or_exhausted (list[int]): List of eliminated candidate indices so far.
+            mutant_tiebreak_record (dict[int, tuple[list[int], int, int]]): Tiebreak record for each round.
+
+        Returns:
+            tuple: (index of eliminated candidate, updated state tuple)
```
cdonnay, line 396: Match the return typing you have above and describe what is being returned.

src/votekit/elections/election_types/ranking/stv.py
```diff
@@ -16,6 +17,1011 @@
...
+            np.inf,
+            tallies,
+        )  # we must do this because sometimes there are cands with 0 FPV who have yet to be eliminated
+        if (
+            np.count_nonzero(masked_tallies == np.min(masked_tallies)) > 1
+        ):
+            potential_losers: list[int] = (
+                np.where(masked_tallies == masked_tallies.min())[0].astype(int).tolist()
+            )
+            L, mutant_tiebreak_record = self.__run_loser_tiebreak(
```
cdonnay, line 409:
```suggestion
            loser_idx, mutant_tiebreak_record = self.__run_loser_tiebreak(
```

src/votekit/elections/election_types/ranking/stv.py
```diff
@@ -16,6 +17,1011 @@
...
+                potential_losers, turn, mutant_tiebreak_record
+            )
+        else:
+            L = int(np.argmin(masked_tallies))
```
cdonnay, line 413:
```suggestion
            loser_idx = int(np.argmin(masked_tallies))
```

src/votekit/elections/election_types/ranking/stv.py
```diff
@@ -16,6 +17,1011 @@
...
+        Args:
+            tallies (np.ndarray): Current tallies for each candidate.
+            turn (int): The current round number.
+            mutant_bool_ballot_matrix (np.ndarray): Boolean mask for eliminated/elected candidates.
+            mutant_winner_list (list[int]): List of winner candidate indices so far.
+            mutant_eliminated_or_exhausted (list[int]): List of eliminated/elected candidate indices so far.
+            mutant_tiebreak_record (dict[int, tuple[list[int], int, int]]): Tiebreak record for each round.
+
+        Returns:
+            tuple: (list of elected candidate indices, updated state tuple)
```
cdonnay, line 454: More specific

src/votekit/elections/election_types/ranking/stv.py
```diff
@@ -16,6 +17,1011 @@
...
+        if self.simultaneous:
+            winners_temp = np.where(tallies >= self.threshold)[0]
+            winners_temp = winners_temp[np.argsort(-tallies[winners_temp])]
+            winners = winners_temp.tolist()
+        else:
+            if np.count_nonzero(tallies == np.max(tallies)) > 1:
+                potential_winners: list[int] = (
+                    np.where(tallies == tallies.max())[0].astype(int).tolist()
+                )
+                w, mutant_tiebreak_record = self.__run_winner_tiebreak(
```
cdonnay, line 465:
```suggestion
                winner_idx, mutant_tiebreak_record = self.__run_winner_tiebreak(
```

src/votekit/elections/election_types/ranking/stv.py
```diff
@@ -16,6 +17,1011 @@
...
+            tied_winners (list[int]): List of candidate indices that are tied.
+            turn (int): The current round number.
+            mutant_tiebreak_record (dict[int, tuple[list[int], int, int]]): Tiebreak record for each round.
+
+        Returns:
+            tuple: (index of new winner, updated tiebreak record)
+        """
+        packaged_tie = frozenset([self.candidates[w] for w in tied_winners])
+        if self._winner_tiebreak == "first_place":
+            W, packaged_ranking = self.__fpv_tiebreak(tied_winners, "winner")
```
cdonnay, line 503: Rename all single letter variable names

src/votekit/elections/election_types/ranking/stv.py
```diff
@@ -16,6 +17,1011 @@
...
+    ) -> tuple[
+        list[np.ndarray],
+        list[tuple[int, list[int], np.ndarray, str]],
+        dict[int, dict[frozenset[str], tuple[frozenset[str], ...]]],
+    ]:
+        """
+        This runs the STV algorithm.
+
+        Args:
+            ballot_matrix (np.ndarray[np.int8]): Matrix where each row is a ballot, each column is a ranking.
```
cdonnay, line 578:
```suggestion
            ballot_matrix (np.ndarray): Matrix where each row is a ballot, each column is a ranking.
```

cdonnay: Also convert to NDArray type setting

src/votekit/elections/election_types/ranking/stv.py
cdonnay: condense_profile does not group ballots by weight; so you only need this if your ballots end up with extra empty frozensets in them. If you want to group ballots by weight, you can call the `group_ballots` method of prefprofile.

tests/elections/election_types/ranking/test_fastSTV.py
```diff
@@ -0,0 +1,439 @@
...
+        candidates=["A", "B", "C", "D", "E"],
+    )
+
+    assert STV(profile, m=3).get_elected() == ({"C"}, {"A"}, {"B"})
+
+
+def test_random_transfers():
+    # in the below profile, B always wins with Cambridge-styled random transfers,
+    # but C would always win with fractional transfers, and wins with probability P > 1 - (1/2)**49 with the "fractional_random" method
+    why_cambridge_is_evil = PreferenceProfile(
```
cdonnay, line 366: change name

tests/elections/election_types/ranking/test_fastSTV.py
```diff
@@ -0,0 +1,439 @@
...
+            Ballot(ranking=(frozenset({"D"}),), weight=73),
+        ),
+        candidates=("A", "B", "C", "D"),
+    )
+    assert STV(profile, m=3, simultaneous=True).get_elected() == ({"A"}, {"B"}, {"D"})
+    assert STV(profile, m=3, simultaneous=False).get_elected() == ({"A"}, {"B"}, {"C"})
+
+
+def test_borda_tiebreak():
+    winning_earlier_is_not_better = PreferenceProfile(
```
cdonnay, line 411: change name

src/votekit/elections/election_types/ranking/stv.py
```diff
@@ -16,6 +17,1011 @@
...
+            tuple[list[np.ndarray], list[tuple[int, list[int], np.ndarray, str]], dict[int, dict[frozenset[str], tuple[frozenset[str], ...]]]]:
+                The tally record is a list with one array per round;
+                    each array counts the first-place votes for the remaining candidates.
+                The play-by-play logs some information for the public methods:
+                    - turn number
+                    - list of candidates elected or eliminated this turn
+                    - weight vector at this turn, if the turn was an election
+                    - turn type: 'election', 'elimination', or 'default'
+                The tiebreak record is a dictionary mapping turn number to a tuple of
+                    (potential candidates involved in tiebreak, chosen candidate, tiebreak type).
```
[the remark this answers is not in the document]
EdouardHeitzmann: not all integers show up in this dict! I could have some entries of the list be empty dicts, but I think it helps the make_election_states method read clearer if I don't have it check for empty dicts.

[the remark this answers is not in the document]
EdouardHeitzmann: As in, only the rounds where a tiebreak was actually necessary show up in the keys of tiebreak_record.

cdonnay (approved):
WOO

PR #305, opened 2025-09-30
Fix cumulative bug

peterrrock2:
Fix the problem where we were missing the total point allocation in new version of cumulative (whoops!)


cdonnay (changes requested):
Just add some doc strings and we're set!

Issue #311, opened 2025-10-13
Rounding errors in STV when redistributing votes


timurka43:
The STV method repeatedly introduces rounding errors when a candidate is elected and remaining votes are redistributed.

Consider the case with 
* k=5 seats, 
* party A having 5 candidates and 52 ballots ranking A1 > A2 > A3 > A4 > A5
* party B having 1 candidate and 8 ballots ranking B1
* N= 60 total voters
* Droop quota t = 60/(5+1) + 1 = 11

The result should be that A1, A2, A3, A4 get sequentially elected subtracting 11 votes from 52 ballots 4 times, until there are only 8 ballots left.
Then A5 and B1 should tie for the last seat. In reality, every time 11 votes are subtracted from A's ballots, there is a rounding error and eventually you end up with 8.000...00014 votes for A5 and 8.0 votes for B1. No ties occur.

See attached code for detailed output of the issue

[rounding_error.py](https://github.com/user-attachments/files/22891425/rounding_error.py)

```
from votekit.ballot import Ballot
from votekit.elections import STV
from votekit.pref_profile import PreferenceProfile
import math

m=5
A=52
N=60
# B gets 8 votes
# Droop quote is 11, so 4 candidates from A should be elected and then 
# the 5th candidate from A should tie with B's only candidate.

#Issue: Rounding errors occur every time the droop quota of 11 is subtracted from A's votes.


# Create candidate names
candidates_A_list = [f"A{i+1}" for i in range(5)]
candidates_B_list = [f"B{i+1}" for i in range(1)]

# Create ballots for party A
ballotA = Ballot(ranking=[{candidate} for candidate in candidates_A_list])
# Create ballots for party B  
ballotB = Ballot(ranking=[{candidate} for candidate in candidates_B_list])

# Create ballot list
ballots = []
ballots.extend([ballotA] * A)  # A ballots for party A
ballots.extend([ballotB] * (N - A))  # N-A ballots for party B

# Create preference profile
candidates = candidates_B_list + candidates_A_list
profile = PreferenceProfile(ballots=ballots, candidates=candidates)

# Run STV election with deterministic tiebreaking from profile
stv_election = STV(profile=profile, m=m)
    
print("STV Election Results:")
print("=" * 50)
print(stv_election)

print("\nVote counts for all candidates that show rounding errors:")
print("=" * 50)

# Get election states which contains detailed round information
election_states = stv_election.election_states
i = 0
# Print the scores for each round that shows rounding errors
[...]

PR #316, opened 2025-11-07
STV rounding error

cdonnay:
Fix the rounding error identified in #311 . Add a 10 decimal round to the utility function `_score_dict_from_rankings_df_no_ties`. Add tests to catch this rounding.

### DIFF
--- a/src/votekit/utils.py
+++ b/src/votekit/utils.py
@@ -175,7 +175,8 @@ def _score_dict_from_rankings_df_no_ties(
     If it is shorter, we add 0s. Candidates who are not mentioned in any ranking do not appear
     in the dictionary.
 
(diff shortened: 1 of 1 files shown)


src/votekit/utils.py
```diff
@@ -294,7 +316,7 @@ def score_dict_from_score_vector(
                 allocation = sum(local_score_vector) / position_size
 
             for c in s:
-                scores[c] += allocation * wt
+                scores[c] += round(allocation * wt, 4)
```
peterrrock2: I don't think that we want to round here since we round in the buckets. Can you check this?

peterrrock2 (approved):

PR #325, opened 2026-02-11
Plurality veto major rework

ross-i:
Overhauled `PluralityVeto`.
Bug fix: now considers unranked candidates as tied for last place, and thus able to be vetoed.
Various optimizations, most importantly relying on profile.df instead of duplicated Ballot objects to conduct election. For deterministic tiebreak, approximately 50-60x faster than previous implementation (testing on a 3-candidate election with 26k voters; was ~2.8s, now <0.05s). For random tiebreak, only about 8-10x faster.
Refactored significantly for readability. Corrected existing documentation and added thorough documentation and comments.
Updated tests for PluralityVeto.

`m` (number of seats) now defaults to 1.
Now supports `tiebreak` = 'random', 'first_place', 'borda', and 'lex'.
`tiebreak` is no longer permitted to be `None`. The default value is 'first_place', and the backup tiebreak is always 'lex', i.e., lexicographic/alphabetical.
Because PluralityVeto requires tiebreaking at the ballot level (rather than round level), PluralityVeto does not report tiebreaks to ElectionState. However, a new attribute `tiebreak_order` allows users to inspect the tiebreak order when using a non-random tiebreak method.


peterrrock2 (commented):
This is looking great!

Small nit: I am trying to unify the doc string style accoss the repository. Can you modify the doc strings to look more like:

```python
def  foo( arg1: str | None, arg2: int = 3) -> str:
    """
    Brief description (max 100 chars including indents)

    Example: 
        <Stuff goes here. This section is optional>

    Args:
        arg1 (str | None): description

    Returns:
        str: The string "baz"
    """
   ...
```

This is a mild modification of the Google-style doc string and adheres to the PEP 257 recommendation but with the line width modified to 100 chars rather than 80 (better for 16:9 screens IMO)

peterrrock2 (approved):
Looks good!

Issue #328, opened 2026-02-24
Improve documentation on spacial models


peterrrock2:
All of the spacial models have parameters that are undocumented and need to have doc strings added to them. 

-  `clustered_spacial_profile_and_positions_generator`
- `onedim_spacial_profile_generator`
- `spacial_profile_and_positions_generator`

PR #329, opened 2026-02-24
Add missing docstrings for spacial model generators (refs #328)

hayan-gh:
Refs #328

Adds missing parameter docstrings for:
- `onedim_spacial_profile_generator`
- `spacial_profile_and_positions_generator`
- `clustered_spacial_profile_and_positions_generator`

Docstrings follow the modified Google style. No behavior changes.

### DIFF
--- a/src/votekit/ballot_generator/std_generator/spacial.py
+++ b/src/votekit/ballot_generator/std_generator/spacial.py
@@ -207,6 +207,16 @@ def clustered_spacial_profile_and_positions_generator(
     Samples candidate positions first, then samples each voter around the
     candidate they are assigned to in ``number_of_ballots``.
 
+    Example:
+        profile, cand_pos, voter_pos = clustered_spacial_profile_and_positions_generator(
+            number_of_ballots={"A": 40, "B": 35, "C": 25},
+            candidates=["A", "B", "C"],
+            voter_dist=np.random.normal,
+            voter_dist_kwargs={"loc": 0.0, "scale": 0.5, "size": 2},
+            candidate_dist=np.random.uniform,
+            candidate_dist_kwargs={"low": -1.0, "high": 1.0, "size": 2},
+        )
+
     Args:
         number_of_ballots (dict[str, int]): The number of voters attributed
                     to each candidate {candidate string: # voters}.


peterrrock2:
Thanks for working on this! I generally review things on Wednesdays, so I will provide feedback tomorrow.

src/votekit/ballot_generator/std_generator/spacial.py
peterrrock2: Fix styling here, please. Brief description is missing.

If you wouldn't mind adding an example, I think that will help users undedrstand how to change the `voter_dist` and `candidate_dist` arguments.

```python
def  foo( arg1: str | None, arg2: int = 3) -> None:
    """
    Brief description (max 100 chars)

    Example: 
        <Stuff goes here>

    Args:
        arg1 (Optional[str]): description
    """
```

src/votekit/ballot_generator/std_generator/spacial.py
```diff
@@ -90,7 +91,20 @@ def spacial_profile_and_positions_generator(
 
     Args:
         number_of_ballots (int): The number of ballots to generate.
-        by_bloc (bool): Dummy variable from parent class.
+        candidates (list[str]): Candidate names used when building rankings.
+        voter_dist (Callable[..., np.ndarray], optional): Distribution sampler used
+            to draw each voter position. Defaults to ``np.random.uniform``.
+        voter_dist_kwargs (Optional[Dict[str, Any]], optional): Keyword arguments
+            passed to ``voter_dist`` for each sample. Defaults to None.
```
peterrrock2: Please specify internal behaviour when the value is "None" so that users with an LSP don't have to guess (we do non-obvious things for the uniform distribution, for example)

peterrrock2 (changes requested):
This is a good start! Just a couple of things to address in here

peterrrock2 (changes requested):
This is looking great! One last thing to make the doc strings in `spacial_profile_and_posisions_generator` and `clustered_spacial_profile_and_positions_generator` parallel with each other

peterrrock2 (approved):
LGTM! @hayan-gh thanks again for your work on this!

PR #333, opened 2026-02-27
SimultaneousVeto implementation

ross-i:
Implemented the voting rule SimultaneousVeto as described in the 2023 paper "Generalized Veto Core and a Practical Voting Rule with Optimal Metric Distortion" (link: https://arxiv.org/abs/2305.19632).

SimultaneousVeto is an anonymous version of PluralityVeto - instead of querying the voters in a random order for their vetoes, all vetoes are processed simultaneously.

Each candidate begins with a score, then, all at the same time, voters remove score from their least favorite candidates at a constant rate. As candidates' scores reach zero, they are eliminated, and voters that were vetoing them transfer to their next least-favorite candidate. Initializing scores as the number of first-place votes (SimultaneousPluralityVeto) yields a voting rule that guarantees worst-case metric distortion of 3, which is the best a deterministic rule can do. However, this implementation of SimultaneousVeto also permits other methods of score initialization, which may have interesting properties.

NOTE: this PR also includes all the commits from the plurality-veto branch since I rebased simultaneous-veto onto plurality-veto.


peterrrock2 (commented):
This is looking great! If you would like to accept these suggestions, we can get you merged in!

PR #362, opened 2026-04-27
Expose speed controls for STV animations

prismika:


prismika:
This is a feature Moon requested. It exposes two optional arguments that modify the speed of an STV animation:
- animation_duration, which sets the amount of seconds each animation step will take, and
- delay_mult, which scales the pauses between animation steps
(I wasn't sure which branch to make the PR against, so it's against main. Feel free to change it.)
Thanks!

peterrrock2 (approved):
LGTM!

PR #367, opened 2026-05-23
Faster mentions

EdouardHeitzmann:
I got tired of waiting >5 seconds every time I need to compute mentions for large profiles.

The legacy mentions function in `votekit/utils` uses the Profile's `.ballots` property to compute ballots; when this property is not already set, it is computed and cached when `mentions` is called. This is quite slow when we are passing a large profile that was loaded via a cvr loader that relies on a pandas df rather than a list of ballots to create the profile. This PR introduces a new pandas-only way to compute mentions in such a case, which is around 14-20x faster on my machine for the Portland profiles.

When the ballot list *is* already materialized, the legacy `mentions` *are* faster; in such a case the `fast_mentions` function I'm adding defaults back to said legacy function.

I also add a few extra tests for both the old and the new mentions functions to verify intended behavior on profiles with repeated and tied rankings (although the former is probably not the intended use case).

Some things worth considering as part of this PR:
- I kept the old mentions calculator as the one named `mentions` here, and I named my addition `fast_mentions`; we might just want to replace `mentions` completely instead.
- This might be a good opportunity to allow compatibility with `Fraction` outputs, if we care (cf. Issue #358). I am happy to do this if desired (this would involve me chasing down other places mentions are used in votekit and tinkering with the typing).

### DIFF
--- a/src/votekit/utils.py
+++ b/src/votekit/utils.py
@@ -1,9 +1,8 @@
 import math
 import random
-from fractions import Fraction
 from itertools import permutations
 from typing import Literal, Optional, Sequence, Union
-
+from typing import Any, cast
 import numpy as np
 import pandas as pd
 from numpy.typing import NDArray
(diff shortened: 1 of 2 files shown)


EdouardHeitzmann:
Ok well it seems like the 3.11 tests are broken -- I'm not sure if this is my doing (and if so, how I did that). Lmk

peterrrock2:
I'm going to merge this into a staging branch  and then work on getting this into 3.4.1. Thanks for your work on this!

PR #368, opened 2026-06-01
Feat/simplify rank ballot input

### LINKED ISSUE #353: Change the way that users make RankBallot objects
RankBallots should take in as their input type `Sequence[str | Iterable[str]]` so that 

```python
Ballot(ranking=[{"A"}, {"B"}, {"C", "D"}], weight=3)
```

Can be simplified to

```python
Ballot(ranking=["A", "B", {"C", "D"}], weight=3)
```

graceg571:
Closes #353 

Relaxes the input requirements for `RankBallot.ranking` so that singleton positions can be provided as bare candidates instead of singleton sets.

Before:
```python
Ballot(ranking=[{"A"}, {"B"}, {"C", "D"}], weight=3)
```

After:

```python
Ballot(ranking=["A", "B", {"C", "D"}], weight=3)
```

## Implementation
Normalization happens at construction where the input is converted to `type[frozenset[str], ...]` within initialization. Validation of ranking is moved after normalization (includes stripping whitespace) to catch `"~ "`  as candidate input. 

## Test
Added case for mixed rankings. Confirms `["A"]` is the same ballot as `[{"A"}]`

### DIFF
--- a/src/votekit/ballot.py
+++ b/src/votekit/ballot.py
@@ -176,14 +176,22 @@ class RankBallot(Ballot):
     ):
         if scores is not None:
             raise TypeError("Only one of ranking or scores can be provided.")
-        ranking = self._convert_ranking_candidates_to_str_strip_whitespace(ranking)
+        ranking = self._convert_ranking_candidates_to_frozenset_strip_whitespace(ranking)
         self._validate_ranking_candidates(ranking)
         self.ranking = ranking
         super().__init__(weight=weight, voter_set=voter_set)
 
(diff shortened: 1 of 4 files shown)


peterrrock2 (changes requested):
TODO:
- Add in the guard clause we talked about in sync code review
- Add in robustness tests
- Fix the Simultaneous veto tests

peterrrock2 (approved):
Looks good!

peterrrock2 (approved):

PR #369, opened 2026-06-01
Feat/sum mult profiles

### LINKED ISSUE #331: Utility method for summing profiles
Add a utility method for summing a list of profiles. You can add profiles right now, but summing a list is tricky if you don't know python well

graceg571:
Closes #331 

Adds a utility method 'sum_profiles' for summing a list of profiles (RankProfile or ScoreProfile). 

## Implementation
Iterates through the list and accumulates with `+`, delegating the actual combination to the existing `__add__` methods on `RankProfile` and `ScoreProfile`. 

Raises `ValueError` on empty list because the correct empty profile type can't be inferred. Raises `TypeError` if a mix of RankProfiles and ScoreProfiles are given as input. Checks before iterating through list to avoid performing unnecessary adds.

## Tests
Added test cases for:
- `sum_profiles([...])` with an empty list
- `sum_profiles([...])` with a list of 1 profile
- `sum_profiles([...])` with a list of mixed ScoreProfiles and RankProfiles
- `sum_profiles([...])` for list of ScoreProfiles and list of RankProfiles

### DIFF
--- a/src/votekit/pref_profile/utils.py
+++ b/src/votekit/pref_profile/utils.py
@@ -429,6 +429,76 @@ def convert_rank_profile_to_score_profile_via_score_vector(
     )
 
 
+def _sum_rank_profiles(rank_profiles: Sequence[RankProfile]) -> RankProfile:
+    """
+    Helper function for sum_profiles that sums RankProfiles.
+
+    Args:
+        rank_profiles (Sequence[RankProfile]): The RankProfiles to sum.
+    """
+
+    from votekit.pref_profile.pref_profile import RankProfile
+
+    candidates = list(set().union(*[set(profile.candidates) for profile in rank_profiles]))
+    max_ranking_length = max([profile.max_ranking_length for profile in rank_profiles])
+
+    total_dfs = []
+    for p in rank_profiles:
+        curr_df = p.df.copy()
+        if p.max_ranking_length < max_ranking_length:
+            for i in range(p.max_ranking_length, max_ranking_length):
+                curr_df.insert(
+                    len(curr_df.columns),
+                    f"Ranking_{i + 1}",
+                    pd.Series([frozenset("~")] * len(curr_df), dtype=object, index=curr_df.index),
+                )
+        total_dfs.append(curr_df)
+
+    new_df = pd.concat(total_dfs, ignore_index=True)
+    new_df.index.name = "Ballot Index"
+    ranking_cols = [c for c in new_df.columns if "Ranking_" in c]
+    new_df[ranking_cols] = new_df[ranking_cols].astype("object")
+    new_df = new_df[
+        [f"Ranking_{i + 1}" for i in range(max_ranking_length)] + ["Weight", "Voter Set"]
+    ]
+
+    return RankProfile(
+        candidates=candidates,
+        df=new_df,
+        max_ranking_length=max_ranking_length,
+    )
+
+
+def _sum_score_profiles(score_profiles: Sequence[ScoreProfile]) -> ScoreProfile:
+    """
+    Helper function for sum_profiles that sums ScoreProfiles.
+
+    Args:
+        score_profiles (Sequence[ScoreProfile]): The ScoreProfiles to sum.
+    """
+
+    from votekit.pref_profile.pref_profile import ScoreProfile
+
+    total_cand = set().union(*[set(profile.candidates) for profile in score_profiles])
+    total_dfs = []
+    for p in score_profiles:
+        curr_df = p.df.copy()
+        curr_cand = set(p.candidates)
+        for cand in total_cand - curr_cand:
+            curr_df[cand] = [np.nan] * len(curr_df)
+        total_dfs.append(curr_df)
+
+    new_df = pd.concat(total_dfs, ignore_index=True)
+    new_df.index.name = "Ballot Index"
+    new_candidates = sorted(total_cand)
+    new_df = new_df[new_candidates + ["Weight", "Voter Set"]]
+
+    return ScoreProfile(
+        candidates=new_candidates,
+        df=new_df,
+    )
+
+
 def sum_profiles(profiles: Sequence[RankProfile | ScoreProfile]) -> RankProfile | ScoreProfile:
     """
     Combines multiple PreferenceProfiles by combining their ball lists.
@@ -444,6 +514,8 @@ def sum_profiles(profiles: Sequence[RankProfile | ScoreProfile]) -> RankProfile 
         TypeError: All profiles must be of the same type.
     """
 
+    from votekit.pref_profile.pref_profile import RankProfile, ScoreProfile
+
     if len(profiles) == 0:
         raise ValueError("Cannot sum an empty list of profiles.")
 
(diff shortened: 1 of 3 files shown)


EdouardHeitzmann:
Just a thought, but using a python loop to repeatedly call `RankProfile.__add__(...)` can be computationally expensive  below:
```
for profile in profiles[1:]:
        combined_profiles = combined_profiles + profile
```
It could speed things up to copy all the underlying `profile.df`s into a list, then call `pd.concat` once on the whole list once -- it's a quadratic speedup in the number of profiles copied (I think!), since for a list of `n` profiles the first once gets copied `n-1` times, the next one `n-2`, etc under the current setup.

EdouardHeitzmann:
Upon further thought, mimicking the `RankProfile.__add__` logic for the many-profiles version of addition as I suggest above would introduce some maintainability concerns (since the same logic would then exist in two separate places). Maybe a clean fix would be to move out the logic currently in `__add__` into the `sum_profiles` helper you're currently working on, make it compatible with arbitrary-length lists of profiles, and then call `sum_profiles` inside of `__add__`:

```
def __add__(self, other) -> RankProfile:
        """
        Add two PreferenceProfiles by combining their ballot lists.
        """
        if not isinstance(other, RankProfile):
            raise TypeError("Unsupported operand type. Must be an instance of RankProfile.")

        # insert any other validation here

        return sum_profiles([self, other])
```

Sorry to be nosing in to this PR, I think the work you're doing is awesome!

src/votekit/pref_profile/utils.py
```diff
@@ -427,3 +427,107 @@ def convert_rank_profile_to_score_profile_via_score_vector(
...
+        TypeError: All profiles must be of the same type.
+    """
+
+    from votekit.pref_profile.pref_profile import RankProfile, ScoreProfile
+
+    if len(profiles) == 0:
+        raise ValueError("Cannot sum an empty list of profiles.")
+
+    if len(profiles) == 1:
+        return profiles[0]
```
peterrrock2:
```suggestion
def sum_profiles(profiles: Sequence[RankProfile | ScoreProfile]) -> RankProfile | ScoreProfile:
    """
    Combines multiple PreferenceProfiles by combining their ball lists.

    Args:
        profiles (Sequence[PreferenceProfile]): The profiles to sum.

    Returns:
        PreferenceProfile: A new PreferenceProfile object containing the combined preference profile.

    Raises:
        ValueError: Cannot sum an empty list of profiles.
        TypeError: All profiles must be of the same type.
    """

    from votekit.pref_profile.pref_profile import RankProfile, ScoreProfile

    if len(profiles) == 0:
        raise ValueError("Cannot sum an empty list of profiles.")

    if len(profiles) == 1:
        return profiles[0] # Change to copy. Might be worth adding a __copy__ method
```

peterrrock2 (changes requested):
Notes:
- Move to internal methods with guaranteed return types (e.g. `_sum_rank_profiles`)

src/votekit/pref_profile/utils.py
graceg571: Added a copy method to RankProfile and ScoreProfile. Updated the return comment

peterrrock2:
> Upon further thought, mimicking the `RankProfile.__add__` logic for the many-profiles version of addition as I suggest above would introduce some maintainability concerns (since the same logic would then exist in two separate places). Maybe a clean fix would be to move out the logic currently in `__add__` into the `sum_profiles` helper you're currently working on, make it compatible with arbitrary-length lists of profiles, and then call `sum_profiles` inside of `__add__`:
> 
> ```
> def __add__(self, other) -> RankProfile:
>         """
>         Add two PreferenceProfiles by combining their ballot lists.
>         """
>         if not isinstance(other, RankProfile):
>             raise TypeError("Unsupported operand type. Must be an instance of RankProfile.")
> 
>         # insert any other validation here
> 
>         return sum_profiles([self, other])
> ```
> 
> Sorry to be nosing in to this PR, I think the work you're doing is awesome!

No worries @EdouardHeitzmann. Thank you for including your input!

src/votekit/pref_profile/utils.py
```diff
@@ -427,3 +427,143 @@ def convert_rank_profile_to_score_profile_via_score_vector(
...
+        )
+
+    total_cand = set().union(*[set(profile.candidates) for profile in score_profiles])
+    total_dfs = []
+    for p in score_profiles:
+        curr_df = p.df.copy()
+        curr_cand = set(p.candidates)
+        for cand in total_cand - curr_cand:
+            curr_df[cand] = [np.nan] * len(curr_df)
+        total_dfs.append(curr_df)
```
peterrrock2, line 527: We could speed this up a little bit by not copying the dataframes that already contain a score for every candidate in total_cand. Avoids a secondary df copy.

tests/pref_profile/utils/test_sum_profiles.py
```diff
@@ -0,0 +1,213 @@
...
+            RankBallot(ranking=({"A"}, {"B"}, {"C"}), weight=2),
+            RankBallot(ranking=({"A", "B"}, frozenset(), {"D"}), voter_set={"Chris"}),
+            RankBallot(),
+            RankBallot(weight=0),
+        ],
+        candidates=["A", "B", "C", "D"],
+        max_ranking_length=3,
+    )
+    summed_profile = sum_profiles([profile])
+    assert summed_profile == profile
```
peterrrock2, line 74:
```suggestion
    assert summed_profile == profile
    assert id(summed_profile) != id(profile) # make sure that we return a copy
```

tests/pref_profile/utils/test_sum_profiles.py
```diff
@@ -0,0 +1,213 @@
...
+        ballots=[
+            ScoreBallot(scores={"A": 2, "B": 2}, weight=2),
+            ScoreBallot(scores={"A": 2, "C": 2}, voter_set={"Chris"}),
+            ScoreBallot(),
+            ScoreBallot(weight=0),
+        ],
+        candidates=["A", "B", "C", "D"],
+    )
+    summed_profile = sum_profiles([profile])
+    assert summed_profile == profile
```
peterrrock2, line 60:
```suggestion
    assert summed_profile == profile
    assert id(summed_profile) != id(profile)
```

peterrrock2 (changes requested):
Just a couple of quick things on this!

src/votekit/pref_profile/utils.py
graceg571: Did not copy the data frame if they already contain a score for each candidate or the same ranking length for summing ScoreProfiles and RankProfiles respectively.

tests/pref_profile/utils/test_sum_profiles.py
graceg571: Added

tests/pref_profile/utils/test_sum_profiles.py
graceg571: Added

peterrrock2 (approved):
LGTM!

PR #370, opened 2026-06-01
Feat/allow int candidates

graceg571:
# Summary

Addresses #359 

This PR allows integer and string candidates as input to rank and score ballots. The user can now input a ranking or score using candidates "1" and 1, and each will be treated as separate candidates with a warning raised to let the user know that's the case.

`RankProfile` and `ScoreProfile` store an internal df of the ballots using integer IDs to represent candidate sets or candidates respectively. Now the values of the dataframe are integers. This provides a uniform internal representation regardless of candidate type which will be helpful for our eventual move to Rust. If the df attribute is accessed, the internal df will be translated to its candidate sets/names and cached for future references.

# Why

Scottish election data labels their candidates as integers and stores a mapping of their integer IDs to candidate names. Likewise, it is common within math literature to use integers to represent candidates. This change will align better with the Scottish election data and math literature as well as allow users to more easily input candidates to ballots without casting all to strings.

We could have cast all integers to strings or enforced no mixing of candidate types, but decided against it: election result candidate names would be misaligned with the ballots the user created, and with the internal df there is no need to constrain candidate types across an entire ballot or preference profile.

# Changes
- `types.py`: added `Candidate : TypeAlias = str | int` and `CandidateFloatDictLike` alias. All type annotations are updated to use `Candidate` across the codebase.
- `ballot.py`: `RankBallot` and `ScoreBallot` now accept `Candidate` in rankings and scores. Warning raised when string and integer candidates collide (e.g. "1" and 1), treated as separate candidates
- `pref_profile.py`: `RankProfile` and `ScoreProfile` maintain and internal `_df` with the candidate integer IDs as the values or column names respectively. The public `df` property translates back to candidate names and is cached. `candidate_id_map` is the mapping of candidates to their integer IDs. `id_candidate_map` is the mapping of candidate integer IDs to candidates. For `RankProfile`, the `candidate_id_map` is the mapping of candidate sets to their IDs. These maps allow translation both ways. This allows the values of the `RankProfile.df` to be integers instead of frozensets of candidates. For `ScoreProfile`, the `candidate_id_map` is the mapping of candidate names to their IDs because the `ScoreProfile.df` has the candidate names as their columns, not sets. A new csv version is introduced. `,v2` is added to the header row of the csv to indicate this new format. `to_csv()` encodes the candidate type within the candidate mapping row tuples e.g. `"(A:str:0), (1:int:1)"`. `from_csv()` reads these tuples, confirms the csv version, and for v2, uses the type to cast the candidate. 
[...]

PR #373, opened 2026-07-08
Fuzzy Ballot Search of Profile

### LINKED ISSUE #326: Add function that queries profiles for particular ballot types
This is a utility function I just made that should be expanded upon to allow for fuzzy search (A over B anywhere in ballot and the like)

```python
from typing import Sequence, Iterable
import pandas as pd
from votekit import RankProfile


def _token_to_frozenset(token: str | Iterable[str]) -> frozenset[str]:
    """Handles the conversion of strings or Sequences of strings to frozensets
    
    Args:
        token (str | Iterable[str]): The input token to be converted to a frozenset

    Returns:
        frozenset[str]: The converted input token
    """
    if isinstance(token, str):
        return frozenset({token.strip()})
    return frozenset(c.strip() for c in token)


def _normalize_ranking(
    ranking: Sequence[str | Iterable[str]],
    max_len: int,
) -> tuple[frozenset[str], ...]:
    """Turns a partial or incompelte ranking into one with the appropriate length for the df query.

    Args:
        ranking (Sequence[str  |  Iterable[str]]): A sequence of strings indicating the ballot type
            to look for. For example ["A", ("B", "C")] will locate the ballot that has "A" ranked 
            first, "B" and "C" both ranked second, and then ranks no other candidates 
        max_len (int): The maximum length of a ballot (used for padding). 

    Raises:
        ValueError: If the input ranking to normalize and pad is longer than what is allowed by
            by max_len

    Returns:
        tuple[frozenset[str], ...]: A converted ranking compatible with VoteKit dataframe queries.
    """
    # VoteKit uses "~" as a reserved placeholder for the end of the ballot
    pad = frozenset({"~"})
    normalized = tuple(_token_to_frozenset(t) for t in ranking)
    if len(normalized) > max_len:
        raise ValueError(f"Ranking has length {len(normalized)} > max_ranking_length {max_len}.")
    return normalized + (pad,) * (max_len - len(normalized))


def find_weight_for_ballot_type(
    ranking: Sequence[str | Iterable[str]],
    profile: RankProfile,
) -> float:
    """Finds the total weight assigned to all ballots with the input ranking.

    Example:
        from votekit import RankBallot, RankProfile

        ballots = [
            RankBallot(ranking=[{"A"}, {"B"}], weight=3),
            RankBallot(ranking=[{"C"}, {"A"}], weight=2),
            RankBallot(ranking=[{"C"}, {"A"}, {"B"}], weight=1),
        ]
        prof = RankProfile(ballots=ballots)
        print(find_weight_for_ballot_type(["C", "A"], prof) ) # prints '2.0'
    
    Args:
        ranking (Sequence[str  |  Iterable[str]]): A sequence of strings indicating the ballot type
            to look for. For example ["A", ("B", "C")] will locate the ballot that has "A" ranked 
            first, "B" and "C" both ranked second, and then ranks no other candidates 
        profile (RankProfile), the profile to query.
    
        
    Returns: 
        float: The total weight assigned to all ballots of the queried type
    """
    # RankProfile.df has columns Ranking_1..Ranking_k, plus Voter Set and Weight.
    max_len = profile.max_ranking_length

    target = _normalize_ranking(ranking, max_len)

    local_dict = {f"r{i+1}": fs for i, fs in enumerate(target)}
    ranking_query = " & ".join([f"Ranking_{i+1} == @r{i+1}" for i in range(max_len)])

    return float(profile.df.query(ranking_query, local_dict=local_dict)["Weight"].sum())
```

graceg571:
# Summary

Addresses #326

This PR adds `search_profile_for_rank_pattern()`, a fuzzy ballot search that returns all rows of a `RankProfile`'s df whose ballots are consistent with a given query. Unlike an exact match, a ballot only needs to satisfy the relations in the query.   Two types of queries can be given: a strict rank order query via ``ranking_query`` or a candidate clone query with a max separation distance via ``max_cand_pair_dist``.  At least one of both arguments can be given.

`ranking_query` specifies a strict order of preference for candidates or candidate sets/tuples. A ballot satisfies the `ranking_query` if it has all the candidates ranked in the order specified.
Each element of the `ranking_query` specifies one ordered slot in the pattern:
- Candidate (str or int): a single candidate that must appear before the candidate(s) in the next slot.
- set of Candidates: a tie (or single candidate) that must occupy a ranking position exactly — the ballot must have that exact set at one position.
- tuple of Candidates/sets: a flexible slot satisfied by any one of the listed candidates or sets appearing before the next slot and/or after the previous slot.

`max_cand_pair_dist` maps candidate pairs to a maximum number of rank slots allowed between them (0 = directly adjacent and includes ties). When multiple candidate pairs are given in ``max_cand_pair_dist``, a ballot satisfies the query if it has all candidate pairs present within their maximum distances. `max_cand_pair_dist` can be used to search for candidate clones.

One optional argument loosens the match: `include_unranked` treats unranked candidates as tied at the end of a ballot (not applied to set slots, since set notation enforces a strict match).

This branch is based on `feat/allow-int-candidates` (#370) and should be merged after because this PR builds on allowing int candidates and uses the internal df representation `profile._df`.

# Why

We often need to pull out ballots of a particular type (e.g. candidate "A" comes before candidate "B") rather than ballots exactly equal to a full ranking. Exact match pulls one specific ballot type. A pattern query pulls the whole family of ballot types that match that pattern.

The implementation leans on #370's id representation. Because the internal `_df` stores candidate sets as integer IDs, each query slot's positions across all ballots can be computed as a boolean matrix, and the pattern reduces to comparing positions between adjacent slots. This also means there is no special handling of string, integer, or mixed candidates.

# Changes

- `utils.py` moved to `utils/` module: existing code moved unchanged to `utils/common_utils.py` and tests moved to `tests/utils/test_common_utils.py` and `tests/utils/test_search_utils.py`
- `utils/search_utils.py`: `search_profile_for_rank_pattern(profile, ranking_query, max_separation_dist, include_unranked)` plus helpers:
  - `_validate_ranking_query()`: type-checks each slot (rejects frozensets, invalid set/tuple contents) and confirms all queried candidates exist in the profile.
  - `_validate_max_cand_dist_dict()`: keys must be candidate-pair tuples referencing adjacent slots in the query, values must be integer distances. If candidates are apart of a set or tuple, only one candidate of the group needs other be in the key. If multiple candidates from a group are in keys, the min distance of the keys is used. 
  - `_get_candidate_ids()`: translates a candidate or exact set into the profile's candidate set IDs via the profile's `candidate_id_map`. A single candidate matches every set containing it, a set matches only its exact set.
  - `_boolean_matrix()`: builds the boolean matrix of a query slot's locations across all ballot rankings from the internal `_df`. With `include_unranked=True`, a virtual end-of-ballot column is appended and used only for ballots where the slot's candidates never appear. Only one end of ballot is marked per ballot.
  - `_get_min_distance_for_cand_pair()`: gets the minimum distance between 2 candidates within the same ballot.
  - `_compare_query_ranks()`: two-pointer sweep over the `np.where` coordinates of adjacent slots' matrices from `ranking_query`, checking rank order and ANDing the ballot mask down the rest of the query.
  - `_compare_candidate_pair_ranks()`: two-pointer sweep over the `np.where` coordinates of candidate pairs' matrices from `max_cand_pair_dist`, checking the pair's max distance separated and ANDing the ballot mask down the rest of the query.
  - 
# Testing

- `test_search_utils.py`: new tests covering a basic candidate pattern, set slots (exact tie match), tuple slots (or-condition), `max_cand_pair_dist`, `include_unranked`, singleton set queries, and an empty-result query. Error-path tests for non-RankProfile input, malformed ranking queries, candidates not in the profile, and invalid `max_cand_pair_dist` keys/values. Mixed str/int candidates, tied positions, and duplicate candidates within a ballot are exercised throughout.


src/votekit/utils/search.py
```diff
@@ -0,0 +1,406 @@
...
+
+
+def _validate_ranking_query(ranking_query: Sequence[Candidate | tuple | set], profile: RankProfile):
+    """
+    Validates the ranking_query is formatted correctly and contains candidates found in the profile.
+
+    Args:
+        ranking_query (Sequence[Candidate | tuple | set]): Ranking pattern to query profile.
+            Candidates can be strings, integers, or mix of both.
+        profile (RankProfile): need profile.candidates, the candidates in profile to query.
```
peterrrock2: Add docs on what makes for a valid ranking query. Also in the public doc string

src/votekit/utils/search.py
```diff
@@ -0,0 +1,406 @@
...
+                        f"Tuple elements must be Candidate or set, got {elm}, {type(elm)}."
+                    )
+        else:
+            raise TypeError(
+                "Ranking query must be a sequence of candidates, tuples, or sets of candidates, got"
+                f" {item}, {type(item)}."
+            )
+    query_candidates = []
+    for item in ranking_query:
+        if isinstance(item, Candidate | set):
```
peterrrock2:
```suggestion
        if isinstance(item, (Candidate, set)):
```
IIRC  the `A | B` syntax has some weird edge cases.

src/votekit/utils/search.py
```diff
@@ -0,0 +1,406 @@
...
+
+    Args:
+        max_separation_dist (dict[tuple, int]): Mapping of candidate pairs to max distances.
+        ranking_query (Sequence[Candidate | tuple | set]): Ranking pattern to query profile.
+
+    Raises:
+        TypeError: max_separation_dist's keys must be a tuple of candidate pairs and values must be
+            integer distances
+        ValueError: The candidate pairs of max_separation_dist's keys must specify candidates that
+            are adjacent to each other in the ranking_query
```
peterrrock2: A little more documentation here, please

src/votekit/utils/search_utils.py
```diff
@@ -0,0 +1,492 @@
...
+            j += 1
+        else:  # a_ballot == b_ballot
+            if a_rank >= b_rank:
+                j += 1  # shift b_rank to the right, keep a_rank still
+            else:  # a ranks before b
+                query_pair_ballots_mask[a_ballot] = True
+                while i < num_query_a_ballots and query_a_ballots[i] == a_ballot:
+                    i += 1  # move pointer to the next unique a_ballot
+                while j < num_query_b_ballots and query_b_ballots[j] == a_ballot:
+                    j += 1  # move pointer to the next unique b_ballot
```
peterrrock2: Maybe change "i" and "j" to "a_where_pair_idx" and "b_where_pair_idx" or something like this

peterrrock2: Also would be a good idea to add signpost comments for this section / pull dense logic out

src/votekit/utils/search_utils.py
```diff
@@ -0,0 +1,492 @@
...
+        b_ballot, b_rank = cand_b_ballots[j], cand_b_ranks[j]
+
+        if a_ballot < b_ballot:
+            i += 1
+        elif a_ballot > b_ballot:
+            j += 1
+        else:  # a_ballot == b_ballot
+            # move to the last rank position of the candidate ranked above the other within a ballot
+            # to minimize distance between cand_a and cand_b
+            if a_rank >= b_rank:
```
peterrrock2: If `a_rank == b_rank` then you are done for this ballot (this is the min distance possible -- tie)

src/votekit/utils/search_utils.py
```diff
@@ -0,0 +1,492 @@
...
+                while (j + 1) < num_cand_b_ballots and cand_b_ballots[j + 1] == a_ballot:
+                    j += 1
+                b_rank = cand_b_ranks[j]
+            else:
+                while (i + 1) < num_cand_a_ballots and cand_a_ballots[i + 1] == a_ballot:
+                    i += 1
+                a_rank = cand_a_ranks[i]
```
peterrrock2: Counterexample [A, X, X, B, A, X, A] will fail for max sep 1 when it should succeed

peterrrock2: Check with test first

src/votekit/utils/search_utils.py
```diff
@@ -0,0 +1,492 @@
...
+        include_unranked (bool): Determines whether unranked candidates are included in query.
+            If True, unranked candidates are considered tied at the end of a ballot.
+            Query slot sets will not include unranked locations because set notation
+            enforces a strict match. Unranked candidates will be excluded by default.
+
+    Returns:
+        (np.array): Boolean matrix where candidate's locations in the _df rankings are marked as
+            True
+
+    """
```
peterrrock2, line 396: Flush this out with example of how unranked works with short ballots

peterrrock2 (changes requested):
.... I forgot to hit the button earlier

src/votekit/utils/search_utils.py
graceg571: Now I compute the minimum distance between a candidate pair within the same ballot and check against that candidate pair's max distance. This will return 0 for a tie.

src/votekit/utils/search_utils.py
graceg571: Added counterexample to tests. Compute the minimum distance between the candidate pair within the same ballot. No more shifting of their pointers to try to get the minimum distance.

src/votekit/utils/search_utils.py
graceg571: Added an example and description of how unranked works with short ballot versus a complete ballot.

src/votekit/utils/search_utils.py
graceg571: Updated "i" to "query/cand_a_where_idx" and "j" to "query/cand_b_where_idx". Moved parts of the code into their own functions to improve readability.

src/votekit/utils/search.py
graceg571: Added documentation on what a ranking_query and its elements represent.

src/votekit/utils/search.py
graceg571: Added more documentation on what makes a valid `max_cand_pair_dist`.

tests/utils/test_search_utils.py
```diff
@@ -0,0 +1,817 @@
...
+        max_ranking_length=4,
+    )
+    profile_match_dist_0 = RankProfile(
+        ballots=(RankBallot(ranking=[{"A", "B"}, {"C"}, 1], weight=1.0),),
+        candidates=("A", "B", "C", "D", 1, 2),
+        max_ranking_length=4,
+    )
+
+    max_sep_dist_0 = {("A", "B"): 0, ("B", "C"): 0}
+    print(search_profile_for_rank_pattern(profile, max_cand_pair_dist=max_sep_dist_0))
```
peterrrock2, line 382:
```suggestion
```

src/votekit/utils/search_utils.py
```diff
@@ -0,0 +1,759 @@
...
+    Returns:
+        (int | None): index into the array of ballot indices where candidate is present and ballot
+            is True from mask. If no ballots with the candidate are true ballots or there are no
+            true ballots, then none is returned.
+
+
+    """
+    cand_curr_ballot = cand_where_ballots[cand_where_idx]
+    num_cand_where_ballots = len(cand_where_ballots)
+    true_ballot_set = set(ballots_where_true)
```
peterrrock2: Not a correctness bug, but I am suspicious of this line. This helper is called twice in the compare function, so creating a set feels really costly

peterrrock2: My suspicion is increasing. We spend practically all of our time in this section:

Image

peterrrock2 (approved):
I think that this is all correct, but I think that we can improve the speed by flipping the primary axis of retention / inquery.

peterrrock2 (approved):
Looks good!

PR #377, opened 2026-08-05
Feat/fraction support

peterrrock2:
## Summary

Restores Fraction ballot-weight support and adds opt-in exact arithmetic to STV, IRV, and Sequential RCV elections.

### Additional Notes

- Support for other classes and election methods is planned in the future. 
- CSV storage through `to_csv` with fractionally-weighted profiles not supported at this time (planned).
- Revisits #311

## Why

Converting rational weights to floats can lose precision, especially for large or closely spaced values. That precision loss can change quotas, candidate rankings, tiebreaks, and election results.

Exact mode preserves rational arithmetic when correctness is more important than the performance of the existing float path. Float arithmetic remains the default for compatibility and speed.

## Changes

- Preserve Fraction weights through ballots, profiles, election states, profile replay, and fractional transfers.
- Add exact=True to STV, IRV, and Sequential RCV for exact tallies, thresholds, and tiebreak scores.
- Add exact-arithmetic helpers and coverage for large values, transfers, replay, tiebreaks, and serialization boundaries.

## Testing

- [x] Relevant checks pass locally
- [x] Tests were added or updated for behavior changes
- [x] Documentation was updated for public API or user-facing changes

## Reviewer Notes

- Exact arithmetic is opt-in. Existing elections continue using the float path by default.
- IRV and Sequential RCV do not perform fractional surplus reweighting, but exact mode still preserves their rational tallies, thresholds, and tiebreak scores.
- Non-integral float weights are rejected in exact mode because their intended rational value is ambiguous.
- NumPy-backed fast election implementations remain on their existing float path.
- Most sensitive file in this PR is `stv.py`. Most other changes are argument threading / type fixes.

### DIFF
--- a/src/votekit/elections/election_types/ranking/stv/exact_fraction_ops.py
+++ b/src/votekit/elections/election_types/ranking/stv/exact_fraction_ops.py
@@ -10,6 +10,10 @@ def to_exact_fraction_weight(value: object) -> Fraction:
     """
     Convert an exact STV ballot weight to a nonnegative ``Fraction``.
 
+    Warning:
+        Exact STV operations assume every ranking position contains at most one candidate. This
+        helper does not validate that profile-level precondition.
+
     Args:
         value (object): Ballot weight to convert. Integral real values and ``Fraction`` values
             are accepted.
@@ -52,6 +56,10 @@ def convert_profile_weights(
     """
     Return a copy of a rank profile with converted ballot weights.
 
+    Warning:
+        Exact STV operations assume every ranking position contains at most one candidate. This
+        helper does not validate that precondition.
+
     Args:
         profile (RankProfile): Profile whose weights are converted.
         converter (Callable[[Any], Numeric]): Function applied to each ballot weight.
@@ -73,6 +81,10 @@ def convert_profile_weights(
 def exact_first_place_votes(profile: RankProfile) -> dict[Candidate, Fraction]:
     """
     Compute exact first-place totals for an exact STV profile.
+
+    Warning:
+        The profile must already have passed STV validation. Tied ranking positions are not
+        supported.
 
     Args:
         profile (RankProfile): Profile with singleton rankings and ``Fraction`` weights.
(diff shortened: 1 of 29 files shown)


src/votekit/elections/election_types/ranking/stv/exact_fraction_ops.py
```diff
@@ -0,0 +1,120 @@
...
+    scores = {candidate: Fraction(0) for candidate in profile.candidates_cast}
+    if profile._df.empty:
+        return scores
+
+    for candidate_id, weight in zip(profile._df["Ranking_1"], profile._df["Weight"]):
+        if candidate_id == -1:
+            continue
+        candidate_set = profile.id_candidate_map[candidate_id]
+        assert len(candidate_set) == 1
+        candidate = next(iter(candidate_set))
```
graceg571, line 104: Is there any check that the profile given only has singleton rankings, no ties?

src/votekit/elections/election_types/ranking/stv/exact_fraction_ops.py
```diff
@@ -0,0 +1,120 @@
...
+    Args:
+        profile (RankProfile): Profile with singleton rankings and ``Fraction`` weights.
+
+    Returns:
+        dict[Candidate, Fraction]: Exact Borda score for every candidate.
+    """
+    scores = {candidate: Fraction(0) for candidate in profile.candidates_cast}
+    assert profile.max_ranking_length is not None
+
+    for ballot in profile.ballots:
```
graceg571: `exact_borda_scores` uses `profile.ballots` instead of the internal `profile._df` while `exact_first_place_votes` uses the internal df. Could be a speed improvement?

graceg571 (approved):
Looks good to me!

One question about `exact_borda_scores` constraint and using the `ballots` attribute versus `_df`

src/votekit/elections/election_types/ranking/stv/exact_fraction_ops.py
peterrrock2: Ah, yep. I had a note go to back and fix that once I verified all the tests and then I forgot to. Good catch!

src/votekit/elections/election_types/ranking/stv/exact_fraction_ops.py
peterrrock2: There is a check in the `__init__` of STV, so there is an implicit assumption here that the caller has done the necessary pre-processing. I'll add that to the doc string.

PR #383, opened 2026-08-06
feat: truncate ranked ballots at sentinel candidates

aryansk:
## Summary

- add `truncate_rank_profile` for CVR markers such as `overvote` and `undervote`
- replace the matching rank and all lower ranks with trailing placeholders while preserving `CleanedRankProfile` metadata
- cover truncation, retention options, and invalid profile types with regression tests

Fixes #380

## Validation

- `uv run pytest -o addopts="" tests -q` — 1,607 passed, 22 skipped
- `uv run ty check src tests`
- `uv run ruff check ...`
- `git diff --check`

Assisted-by: Codex:gpt-5

PR #387, opened 2026-08-11
feat: add profile subtraction utility

### LINKED ISSUE #384: Utility method for subtracting profiles
Profiles can currently be combined using `+` operator or `sum_profiles`. With the new ballot pattern search function `search_profile_for_rank_pattern`,  we can retrieve ballots where a candidate pair is separated at most x rank positions. To support the inverse query, separation of at least x, we need profile subtraction.

Profiles with same ballot rankings but differing weights could subtract the weights and leave the voter set untouched.

aryansk:
## Problem

Profiles can be combined with `+` / `sum_profiles`, but there is no inverse operation. With `search_profile_for_rank_pattern` available, queries like "ballots where a candidate pair is separated by at least x ranks" need profile subtraction.

## Change

- Add `subtract_profiles(minuend, subtrahend)` in `pref_profile/utils.py`.
- Add `__sub__` to `RankProfile` and `ScoreProfile`.
- Ballot weights are subtracted for identical rankings (or scores); the minuend's voter sets are retained.
- A `ValueError` is raised if any weight would become negative.

## Testing

- New test file `tests/pref_profile/utils/test_subtract_profiles.py` (9 tests) covering weight subtraction, voter-set retention, negative-weight errors, mixed-length rankings, mixed types, and the standalone function.
- Full `tests/pref_profile/` suite passes (170 tests).

Closes #384

### DIFF
--- a/src/votekit/pref_profile/utils.py
+++ b/src/votekit/pref_profile/utils.py
@@ -574,3 +574,159 @@ def sum_profiles(profiles: Sequence[PreferenceProfile]) -> PreferenceProfile:
             f"Cannot sum profiles of type {type(profiles[0]).__name__}. "
             "List can only contain RankProfiles or ScoreProfiles."
         )
+
+
+def subtract_profiles(
+    minuend: PreferenceProfile, subtrahend: PreferenceProfile
+) -> PreferenceProfile:
+    """
+    Subtract the ballot weights of one profile from another, matched by
+    identical ballot rankings (or scores).
+
+    Profiles with the same ballot rankings but differing weights have their
+    weights subtracted, leaving the voter set of the minuend untouched.
+
+    Args:
+        minuend (PreferenceProfile): The profile being subtracted from.
+        subtrahend (PreferenceProfile): The profile whose weights are subtracted.
+
+    Returns:
+        PreferenceProfile: A new profile with the subtracted ballot weights.
+
+    Raises:
+        TypeError: Both profiles must be of the same type, RankProfile or ScoreProfile.
+        ValueError: A ballot weight would become negative.
+    """
+
+    from votekit.pref_profile.pref_profile import RankProfile, ScoreProfile
+
+    if type(minuend) is not type(subtrahend):
+        raise TypeError(
+            f"Both profiles must be of the same type. "
+            f"Got {type(minuend).__name__} and {type(subtrahend).__name__}."
+        )
+
+    if isinstance(minuend, RankProfile):
+        return _subtract_rank_profiles(minuend, subtrahend)
+
+    if isinstance(minuend, ScoreProfile):
+        return _subtract_score_profiles(minuend, subtrahend)
+
+    raise TypeError(
+        f"Cannot subtract profiles of type {type(minuend).__name__}. "
+        "Only RankProfiles or ScoreProfiles are supported."
+    )
+
+
+def _subtract_rank_profiles(minuend: "RankProfile", subtrahend: "RankProfile") -> "RankProfile":
+    """Helper function for subtract_profiles that subtracts RankProfiles."""
+
+    from votekit.pref_profile.pref_profile import RankProfile
+
+    candidates = list(set().union(*[set(profile.candidates) for profile in [minuend, subtrahend]]))
+    max_ranking_length = max(
+        [
+            profile.max_ranking_length
+            for profile in [minuend, subtrahend]
+            if profile.max_ranking_length is not None
+        ]
+    )
+
+    # Pad both profiles to the same ranking length
+    padded_dfs = []
+    for profile in [minuend, subtrahend]:
+        assert profile.max_ranking_length is not None
+        curr_df = profile.df.copy()
+        for i in range(profile.max_ranking_length, max_ranking_length):
+            curr_df.insert(
+                len(curr_df.columns),
+                f"Ranking_{i + 1}",
+                pd.Series([frozenset("~")] * len(curr_df), dtype=object, index=curr_df.index),
+            )
+        padded_dfs.append(curr_df)
+
+    minuend_df, subtrahend_df = padded_dfs
+    ranking_cols = [f"Ranking_{i + 1}" for i in range(max_ranking_length)]
+
+    # Align by ranking and subtract weights
+    subtrahend_weights = subtrahend_df.groupby(ranking_cols, dropna=False)["Weight"].sum()
+    minuend_grouped = minuend_df.groupby(ranking_cols, dropna=False)
+    new_rows = []
+    for key, group in minuend_grouped:
+        key = key[0] if isinstance(key, tuple) and len(key) == 1 else key
+        weight = group["Weight"].sum() - subtrahend_weights.get(key, 0)
+        if weight < 0:
+            raise ValueError(
+                f"Cannot subtract profiles: ballot weight would become negative for ranking {key}."
+            )
+        voter_set = set().union(*group["Voter Set"])
+        ranking = {
+            col: key[i] if i < len(key) else frozenset("~") for i, col in enumerate(ranking_cols)
+        }
+        new_rows.append(
+            {
+                **ranking,
+                "Weight": weight,
+                "Voter Set": voter_set,
+            }
+        )
+
+    # Ballots present only in the subtrahend do not appear in the result
+    if new_rows:
+        new_df = pd.DataFrame(new_rows)[ranking_cols + ["Weight", "Voter Set"]]
+    else:
+        new_df = pd.DataFrame(columns=ranking_cols + ["Weight", "Voter Set"], dtype=object)
+    new_df.index.name = "Ballot Index"
+
+    return RankProfile(
+        candidates=candidates,
+        df=new_df,
+        max_ranking_length=max_ranking_length,
+    )
+
+
+def _subtract_score_profiles(minuend: "ScoreProfile", subtrahend: "ScoreProfile") -> "ScoreProfile":
+    """Helper function for subtract_profiles that subtracts ScoreProfiles."""
+
+    from votekit.pref_profile.pref_profile import ScoreProfile
+
+    total_cand = set().union(*[set(profile.candidates) for profile in [minuend, subtrahend]])
+    minuend_df = minuend.df.copy()
+    subtrahend_df = subtrahend.df.copy()
+    for df in (minuend_df, subtrahend_df):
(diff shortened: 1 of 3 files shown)


src/votekit/pref_profile/utils.py
```diff
@@ -574,3 +574,159 @@ def sum_profiles(profiles: Sequence[PreferenceProfile]) -> PreferenceProfile:
             f"Cannot sum profiles of type {type(profiles[0]).__name__}. "
             "List can only contain RankProfiles or ScoreProfiles."
         )
+
+
+def subtract_profiles(
+    minuend: PreferenceProfile, subtrahend: PreferenceProfile
```
graceg571, line 580: `minuend` and `subtrahend` are not intuitive to those without a math background. `base_profile` and `subtracted_profiles` would be clearer.

graceg571, line 580: Similar to `sum_profiles`, the `subtrahend` could be a list of profiles or a single profile, and the profiles can be summed together prior to subtraction

src/votekit/pref_profile/utils.py
```diff
@@ -574,3 +574,159 @@ def sum_profiles(profiles: Sequence[PreferenceProfile]) -> PreferenceProfile:
...
+        "Only RankProfiles or ScoreProfiles are supported."
+    )
+
+
+def _subtract_rank_profiles(minuend: "RankProfile", subtrahend: "RankProfile") -> "RankProfile":
+    """Helper function for subtract_profiles that subtracts RankProfiles."""
+
+    from votekit.pref_profile.pref_profile import RankProfile
+
+    candidates = list(set().union(*[set(profile.candidates) for profile in [minuend, subtrahend]]))
```
graceg571, line 626: A ballot can only be subtracted if it exists in the `minuend` profile. Therefore, the total list of candidates is always the same as `minuend`'s `candidates`. 

Likewise, we would want to define what we do if a ballot to be subtracted does not exist in the `minuend` profile.

src/votekit/pref_profile/utils.py
```diff
@@ -574,3 +574,159 @@ def sum_profiles(profiles: Sequence[PreferenceProfile]) -> PreferenceProfile:
...
+    max_ranking_length = max(
```
graceg571, line 627: Building off the `candidates` comment, the `max_ranking_length` should be the `minuend` profile as that's the "ground truth" we remove ballots from.

src/votekit/pref_profile/utils.py
```diff
@@ -574,3 +574,159 @@ def sum_profiles(profiles: Sequence[PreferenceProfile]) -> PreferenceProfile:
...
+    for profile in [minuend, subtrahend]:
+        assert profile.max_ranking_length is not None
+        curr_df = profile.df.copy()
+        for i in range(profile.max_ranking_length, max_ranking_length):
+            curr_df.insert(
+                len(curr_df.columns),
+                f"Ranking_{i + 1}",
+                pd.Series([frozenset("~")] * len(curr_df), dtype=object, index=curr_df.index),
+            )
+        padded_dfs.append(curr_df)
```
graceg571, line 646: No need to pad the profiles. We only care if ballots from `subtrahend` profile match ballots from `minuend` profile. Ballots with the same rankings will be equivalent with/without padding.

src/votekit/pref_profile/utils.py
```diff
@@ -574,3 +574,159 @@ def sum_profiles(profiles: Sequence[PreferenceProfile]) -> PreferenceProfile:
...
+
+    minuend_df, subtrahend_df = padded_dfs
+    ranking_cols = [f"Ranking_{i + 1}" for i in range(max_ranking_length)]
+
+    # Align by ranking and subtract weights
+    subtrahend_weights = subtrahend_df.groupby(ranking_cols, dropna=False)["Weight"].sum()
+    minuend_grouped = minuend_df.groupby(ranking_cols, dropna=False)
```
graceg571, line 653: There's a `group_ballots` instance method for rank and score profiles.

src/votekit/pref_profile/utils.py
```diff
@@ -574,3 +574,159 @@ def sum_profiles(profiles: Sequence[PreferenceProfile]) -> PreferenceProfile:
...
+    new_rows = []
+    for key, group in minuend_grouped:
+        key = key[0] if isinstance(key, tuple) and len(key) == 1 else key
+        weight = group["Weight"].sum() - subtrahend_weights.get(key, 0)
+        if weight < 0:
+            raise ValueError(
+                f"Cannot subtract profiles: ballot weight would become negative for ranking {key}."
+            )
```
graceg571, line 661: A negative weight could be an error if the user wants to enforce the `subtrahend` profile being a subset of the `minuend` profile. For instance, if we were subtracting by weight, a negative weight would indicate subtraction has gone beyond the weight allocated to that ballot. For this function, the `subtrahend` does not need to be a subset of the `minuend`, and instead the `subtrahend` is the hand that removes ballots from the `minuend` bag. Based on this analogy, if the hand goes to remove a ballot that no longer exists in the bag, then nothing is removed. Therefore, a ballot's weight should be constrained to no less than 0 and removed from the profile. We can have a parameter to state whether we remove those zero weight ballots and have the default be to remove them.

peterrrock2:
(copy of comment on #383)

Hey @aryansk!

Thank you so much for contributing to VoteKit! It's wonderful to see our work being used by people outside of our community. If you happen to have a moment, I would love to hear more about how you found the package and what you are using it for so that we can better fit the needs of our users.

You may have noticed that we made our latest release of VoteKit around two weeks ago. Since then, I have been preoccupied with some other projects and missed the latest set of PRs that came in. This looks like a great start on this issue, and one of the other maintainers will be adding a review with some additional feedback shortly.

I also saw you mention that you were assisted by Codex when working on this issue. Thank you for including that information. As an academic lab, we value transparency of process, and knowing this sort of thing is genuinely helpful. I think that, until this point, we have not been clear about our preferences for how contributors use AI tools when submitting work for review on this project.

I hope the following makes this clear, but I want to state it explicitly anyway: the recent updates to this policy are not an indictment of any of the work that you have done here. As a maintainer of a small OSS package, it's always heartening to see more people trying to become a part of the community, and I am truly grateful for the work that you have put in.

---
Additional notes (this section is a form that I am including on several recent PRs)

Since there have been so many new contributors in the last couple of weeks (a wonderful problem to have), I thought it important to update the "Use of AI Tools" section of the [Contributing.md](https://github.com/mggg/VoteKit/edit/main/CONTRIBUTING.md) on our main branch to clarify what we currently consider to be best-practices for our lab. 

To be clear, this this is not an indictment of the use of any AI tools or a statement about the work presented here (indeed, it would be a bit illogical to find anyone at issue with this policy before it was explicitly stated!).  I include it here only because this is 1. a recent update in a document people infrequently visit and 2. an effort to clearly communicate our expectations around contributing to this project. 

I'll include the exact snippet here so you don't have to go through the trouble of finding it yourself:

> ## Use of AI tools
>
> Researchers rely on VoteKit and its results, so contributors have a heightened responsibility for
> the correctness and clarity of every change.
>
> - Core implementation and business logic in `src/votekit/` must be written by humans. AI tools may
>   provide inline assistance (e.g., an editor copilot) and may be used to navigate and deepen
>   understanding of the codebase and domain material (e.g., with chat agents). Contributors should
>   not delegate the generation or rewriting of core implementation code to agentic coding tools.
>   For example, Codex GUI/CLI, Claude GUI/CLI, Cursor v3+, and similar tools should not be asked to
>   generate or substantially rewrite that code.
> - Tests may be developed jointly by humans and AI. AI is particularly useful for adversarial
>   testing, such as searching for user inputs that break an implementation, but the person
>   responsible for the pull request must review every AI-generated or AI-assisted test and confirm
>   that it meaningfully tests the intended behavior.
> - Docstrings may be developed jointly by humans and AI. The person responsible for the pull
>   request must read, edit, and review every AI-assisted docstring for accuracy, thoroughness, and
>   compliance with the documentation and formatting guidelines below.
>
> The person responsible for a pull request remains accountable for all of its contents, regardless
> of which tools assisted in preparing it.

---

Thank you again  for your contribution, and I hope that we'll see more from you soon!

-Peter
(Lead Maintainer of VoteKit)

graceg571 (changes requested):
Thanks for contributing this PR! Subtracting profiles can take many different forms: 
- set difference: subtract only rankings and ignore its weigh. If there's a ranking match between the base profile and the subtracted profile(s), the ranking is removed by default.
-  reduce by weight: subtract ballots by weight. The ranking's weight will be reduced by the weight value of the matching ranking in the subtracted profile. An error would be thrown when a ballot does not exist or goes negative weight because weight cannot be reduced from a missing ballot and a negative weight is not a valid weight reduction.
-  "normal" ballot subtraction: the base profile is the bag of ballots to be removed/subtracted by the subtracted profile(s). A ballot in the subtracted profile found in the base profile is subtracted (or removed from the bag) till the weight of the subtracted ranking's weight is consumed or the base ranking's weight is exhausted (i.e. zero weight). So, zero weight rankings are removed from the profile by default.

We would like to start with the last ballot subtraction type, and my review comments reflect that decision. Let me know if you want to keep working on this!

peterrrock2:
@aryansk just following up to see if you saw the above messages. Thanks!

fetachino:
One portability concern: candidates is built with list(set().union(...)), which makes the candidate order nondeterministic. If RankProfile serialization, equality, or downstream display relies on candidate order, subtraction could produce different results between runs. Could this preserve the minuend (or a deterministic merged) candidate order, with a regression test?

graceg571:
> One portability concern: candidates is built with list(set().union(...)), which makes the candidate order nondeterministic. If RankProfile serialization, equality, or downstream display relies on candidate order, subtraction could produce different results between runs. Could this preserve the minuend (or a deterministic merged) candidate order, with a regression test?

Good catch! Since subtraction removes ballots from the minuend, every candidate in the result already appears in the minuend. So, the union is unnecessary. You can use the minuend's candidates directly which preserves its order deterministically and resolves the serialization and display concerns. We keep the minuend's complete candidate list rather than filtering to candidates with surviving ballots because the candidates are important election metadata, and the ballot-level view is captured by `candidates_cast`. 

Equality is unaffected either way because `RankProfile` compares candidates as sets, so order doesn't matter. A regression test would be great to add to ensure the resulting profile's candidates matches the minuend's.

PR #389, opened 2026-08-17
Remove ballots with candidate(s) cleaning function and update `CleanedRankProfile` index tracking

### LINKED ISSUE #381: Remove ranked ballots containing specified candidates
## Summary

Add a profile cleaning function that removes every ballot containing one of a specified set of candidates or CVR markers.

This supports workflows where ballots containing values such as `"overvote"` or an invalid write-in marker should be excluded entirely rather than having the marker removed and the ballot condensed.

## Expected behavior

Given ballots:

```python
RankBallot(ranking=["A", "B"])
RankBallot(ranking=["A", "overvote", "C"])
RankBallot(ranking=["C", "A"])
```

filtering on "overvote" should retain the first and third ballots and remove the second.

Part of #337.

graceg571:
Closes #381 

## Summary 
`remove_ballots_with_cand_rank_profile` is added as a cleaning function to remove ballots containing a certain candidate or one from a set of candidates. Currently, specific candidate(s) can be removed from a ballot, but this supports workflows where ballots with an invalid write-in marker like `"overvote"` or `"undervote"` can be removed entirely.

## Changes
- `remove_ballots_with_cand_rank_profile` is added to `rank_profiles_cleaning.py`. The ballots are removed using the candidates' integer IDs within internal `df`. The removed ballots are treated the same as dropped empty ranking and zero weight ballots where they are gaps in the index when compared to the parent profile.
- `no_rank_altr_idxs` is modified to include ballots where all rank slots are `frozsenset()` or `frozenset("~")`. This accounts for ballots where the cleaning function removed all ranked candidates and replaced them with empty frozensets.
- `unaltr_idxs` are removed from `no_rank_altr_idxs` to correctly distinguish which ballots were already empty and are empty due to cleaning.
- `unaltr_idxs` contains ballots that were dropped but not modified by the cleaning function. Empty ranking ballots and zero weight ballots fall in this bucket. The ballots removed by `remove_ballots_with_cand_rank_profile` are also tracked in `unaltr_idxs`. To determine which ballots were dropped, the index of the resulting cleaned profile can be compared to the index of the original parent profile.
- `is_equiv_to_condensed` updated to return True if no `frozenset()` in ranking or only trailing ones alongside any `frozenset("~")`. 
- `is_equiv_for_remove_and_condense` updated to use `is_equiv_to_condensed`

## Testing
- `test_remove_ballots_with_cand_rank_profile.py` added with tests for ballots with `removed` candidates with and without ties
- `test_clean_ranked_profile.py` and `test_remove_cand_ranked_profile.py` updated to account for the changes to `no_rank_altr_idxs` and `unaltr_idxs` now including originally empty ballots
- `test_condense_ranked_profile` updated to account for changes to `is_equiv_to_condensed` to include originally empty ballots

### DIFF
--- a/src/votekit/cleaning/rank_profiles_cleaning.py
+++ b/src/votekit/cleaning/rank_profiles_cleaning.py
@@ -6,10 +6,10 @@ import pandas as pd
 
 from votekit.pref_profile import (
     CleanedRankProfile,
-    ProfileError,
     RankProfile,
 )
 from votekit.types import Candidate, CandidateList
+from votekit.utils import _validate_candidate_names
 
 
 def _iterate_and_clean_ranking_tuples(
@@ -65,7 +65,7 @@ def clean_rank_profile(
     clean_ranking_func: Callable[[tuple], tuple],
     remove_empty_ballots: bool = True,
     remove_zero_weight_ballots: bool = True,
-    retain_original_candidate_list: bool = True,
+    retain_original_candidate_list: bool = False,
 ) -> CleanedRankProfile:
     """
     Allows user-defined cleaning rules for RankProfile. Input function
@@ -81,16 +81,16 @@ def clean_rank_profile(
             weight as a result of the cleaning. Defaults to True.
         retain_original_candidate_list (bool, optional): Whether or not to use the candidate list
             from the original profile in the new profile. If False, uses only candidates who receive
-            votes. Defaults to True.
+            votes. Defaults to False.
 
     Returns:
         CleanedRankProfile: A cleaned ``RankProfile``.
 
     Raises:
-        ProfileError: Profile must only contain ranked ballots.
+        TypeError: Profile must only contain ranked ballots.
     """
     if not isinstance(profile, RankProfile):
-        raise ProfileError("Profile must be a RankProfile.")
+        raise TypeError("Profile must be a RankProfile.")
 
     (
         cleaned_df,
@@ -187,7 +187,7 @@ def remove_repeat_cands_rank_profile(
         CleanedRankProfile: A cleaned ``RankProfile``.
 
     Raises:
-        ProfileError: Profile must only contain ranked ballots.
+        TypeError: Profile must only contain ranked ballots.
     """
 
     return clean_rank_profile(
@@ -260,7 +260,7 @@ def remove_cand_rank_profile(
         CleanedRankProfile: A cleaned ``RankProfile``.
 
     Raises:
-        ProfileError: Profile must only contain ranked ballots.
+        TypeError: Profile must only contain ranked ballots.
     """
     if isinstance(removed, Candidate):
         removed = [removed]
@@ -531,8 +531,9 @@ def remove_ballots_with_cand_rank_profile(
     """
     Given a ranked profile, remove the ballots that contain the given candidate(s).
 
-    A removed ballot's ranking is considered empty after cleaning and recorded in the
-    ``no_rank_altr_idxs`` of the returned ``CleanedRankProfile``.
+    A removed ballot is not considered altered after cleaning and is accounted for in the difference
+    between the parent profile's index versus the indices of the returned
+    ``CleanedRankProfile``. The same goes for removed empty ranking and zero weight ballots.
 
     Args:
         removed (Candidate | list[Candidate]): Candidate or list of candidates to remove their
@@ -550,21 +551,17 @@ def remove_ballots_with_cand_rank_profile(
             candidate(s) removed.
 
     Raises:
-        ProfileError: Profile must contain ranked ballots.
+        TypeError: Profile must contain ranked ballots.
         TypeError: Candidates to be removed must be strings or integers. A boolean or float
             candidate of the same value as an integer candidate would result in removing the ballots
             with that integer candidate.
     """
     if not isinstance(profile, RankProfile):
-        raise ProfileError("Profile must be a RankProfile.")
+        raise TypeError("Profile must be a RankProfile.")
 
-    if isinstance(removed, Candidate) and not isinstance(removed, bool):
+    if isinstance(removed, Candidate):
         removed = [removed]
-    elif isinstance(removed, list):
-        if any(not isinstance(cand, (str, int)) or isinstance(cand, bool) for cand in removed):
-            raise TypeError("Candidates must be strings or integers within removed.")
-    else:
-        raise TypeError("removed must be a str/int candidate or a list of candidates.")
+    _validate_candidate_names(removed, attribute="removed")
 
     cand_ids = []
     for cand in removed:
@@ -578,7 +575,6 @@ def remove_ballots_with_cand_rank_profile(
     ranking_cols = [f"Ranking_{i}" for i in range(1, profile.max_ranking_length + 1)]
     ballots_to_remove = profile._df[ranking_cols].isin(cand_ids).any(axis=1)
     cleaned_df = profile.df[~ballots_to_remove]
-    removed_ballot_idxs = set(profile.df[ballots_to_remove].index)
 
     if remove_empty_ballots:
         mask = cleaned_df[ranking_cols].map(lambda x: x == frozenset({"~"})).all(axis=1)
@@ -593,15 +589,14 @@ def remove_ballots_with_cand_rank_profile(
         else tuple(set(profile.candidates) - set(removed))
     )
 
-    unaltered_idxs = list(cleaned_df.index)
     return CleanedRankProfile(
         df=cleaned_df,
         candidates=candidates,
         max_ranking_length=profile.max_ranking_length,
         parent_profile=profile,
-        df_index_column=unaltered_idxs,
+        df_index_column=list(cleaned_df.index),
         no_wt_altr_idxs=set(),
-        no_rank_altr_idxs=removed_ballot_idxs,
+        no_rank_altr_idxs=set(),
         nonempty_altr_idxs=set(),
-        unaltr_idxs=set(unaltered_idxs),
+        unaltr_idxs=set(profile.df.index),  # ballots are only dropped, never altered
     )
--- a/src/votekit/cleaning/rank_profiles_cleaning.py
+++ b/src/votekit/cleaning/rank_profiles_cleaning.py
@@ -537,8 +537,8 @@ def remove_ballots_with_cand_rank_profile(
     """
     Given a ranked profile, remove the ballots that contain the given candidate(s).
 
-    A removed ballot is considered altered for both weight and ranking. No ballots are altered in
-    place, only removed.
+    A removed ballot's ranking is considered empty after cleaning and recorded in the
+    ``no_rank_altr_idxs`` of the returned ``CleanedRankProfile``.
 
     Args:
         removed (Candidate | list[Candidate]): Candidate or list of candidates to remove their
@@ -584,19 +584,14 @@ def remove_ballots_with_cand_rank_profile(
     ranking_cols = [f"Ranking_{i}" for i in range(1, profile.max_ranking_length + 1)]
     ballots_to_remove = profile._df[ranking_cols].isin(cand_ids).any(axis=1)
     cleaned_df = profile.df[~ballots_to_remove]
-    removed_ballot_idxs = set(profile.df.index[ballots_to_remove])
+    removed_ballot_idxs = set(profile.df[ballots_to_remove].index)
 
-    empty_ballot_idxs: set[int] = set()
     if remove_empty_ballots:
         mask = cleaned_df[ranking_cols].map(lambda x: x == frozenset({"~"})).all(axis=1)
-        empty_ballot_idxs = set(cleaned_df.index[mask])
         cleaned_df = cleaned_df[~mask]
 
-    zero_weight_ballot_idxs: set[int] = set()
     if remove_zero_weight_ballots:
-        mask = cleaned_df["Weight"] > 0
-        zero_weight_ballot_idxs = set(cleaned_df.index[~mask])
-        cleaned_df = cleaned_df[mask]
+        cleaned_df = cleaned_df[cleaned_df["Weight"] > 0]
 
     candidates = (
         profile.candidates
@@ -611,8 +606,8 @@ def remove_ballots_with_cand_rank_profile(
         max_ranking_length=profile.max_ranking_length,
         parent_profile=profile,
         df_index_column=unaltered_idxs,
-        no_wt_altr_idxs=removed_ballot_idxs | zero_weight_ballot_idxs,
-        no_rank_altr_idxs=removed_ballot_idxs | empty_ballot_idxs,
+        no_wt_altr_idxs=set(),
+        no_rank_altr_idxs=removed_ballot_idxs,
         nonempty_altr_idxs=set(),
         unaltr_idxs=set(unaltered_idxs),
     )
--- a/src/votekit/cleaning/rank_profiles_cleaning.py
+++ b/src/votekit/cleaning/rank_profiles_cleaning.py
@@ -40,10 +40,14 @@ def _iterate_and_clean_ranking_tuples(
     cleaned_df[ranking_cols] = pd.DataFrame(cleaned_rows, index=cleaned_df.index)
 
     tilde = frozenset({"~"})
+    empty = frozenset()
     idxs = cleaned_df.index
 
     unaltr_idxs = {idx for idx, (o, c) in zip(idxs, zip(orig_rows, cleaned_rows)) if o == c}
-    no_rank_altr_idxs = {idx for idx, c in zip(idxs, cleaned_rows) if all(x == tilde for x in c)}
+    no_rank_altr_idxs = {
+        idx for idx, c in zip(idxs, cleaned_rows) if all(x == tilde or x == empty for x in c)
+    }
+    no_rank_altr_idxs = no_rank_altr_idxs - unaltr_idxs
     nonempty_altr_idxs = set(idxs) - unaltr_idxs - no_rank_altr_idxs
     no_wt_altr_idxs: set[int] = set()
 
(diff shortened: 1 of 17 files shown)


tests/cleaning/rank_profiles/test_remove_ballots_with_cand_rank_profile.py
peterrrock2, line 1: Fixtures
Test all attrs for CleanProfile independently
Test chaining
Test idempotent

src/votekit/cleaning/rank_profiles_cleaning.py
```diff
@@ -40,10 +40,14 @@ def _iterate_and_clean_ranking_tuples(
...
     tilde = frozenset({"~"})
+    empty = frozenset()
     idxs = cleaned_df.index
 
     unaltr_idxs = {idx for idx, (o, c) in zip(idxs, zip(orig_rows, cleaned_rows)) if o == c}
-    no_rank_altr_idxs = {idx for idx, c in zip(idxs, cleaned_rows) if all(x == tilde for x in c)}
+    no_rank_altr_idxs = {
+        idx for idx, c in zip(idxs, cleaned_rows) if all(x == tilde or x == empty for x in c)
+    }
+    no_rank_altr_idxs = no_rank_altr_idxs - unaltr_idxs
```
peterrrock2, line 50: Document change in PR description, and update the doc string in the CleanProfile class. Please add examples in that doc string so we can more easily determine the meaning of these.

### TODO:

Add issue where we use a dataclass to descriptively partition the possible state space for these alterations -- users should make the decision on what they car about


example

```python

@dataclass(slots = true)
class CleaningIndexDeltas:
    dropped: set or list or tuple --> entire row removed index does not exist in child
    remove_empty_ranking: TypedDict { trailing_only: <object>, internal_only: <object>, both_trailing_and_internal: <object>}
    subset_previous_ballot: TypedDict {strict_subset_no_gaps ([A, B, {C, D}, E] -> [B, {C,D}]), strict_subset_gaps, weak_subset ([A, B, {C,D}, E] -> [B, C]) 
... and so on
```

This is a loose sketch

src/votekit/cleaning/rank_profiles_cleaning.py
```diff
@@ -521,7 +515,93 @@ def remove_and_condense_rank_profile(
...
+    profile: RankProfile,
+    remove_empty_ballots: bool = True,
+    remove_zero_weight_ballots: bool = True,
+    retain_original_candidate_list: bool = False,
+) -> CleanedRankProfile:
+    """
+    Given a ranked profile, remove the ballots that contain the given candidate(s).
+
+    A removed ballot's ranking is considered empty after cleaning and recorded in the
+    ``no_rank_altr_idxs`` of the returned ``CleanedRankProfile``.
```
peterrrock2: The current paradigm (which is not great) records dropped ballots as gaps in the index (compared to the parent profile)

src/votekit/cleaning/rank_profiles_cleaning.py
```diff
@@ -521,7 +515,93 @@ def remove_and_condense_rank_profile(
...
+            candidate(s) removed.
+
+    Raises:
+        ProfileError: Profile must contain ranked ballots.
+        TypeError: Candidates to be removed must be strings or integers. A boolean or float
+            candidate of the same value as an integer candidate would result in removing the ballots
+            with that integer candidate.
+    """
+    if not isinstance(profile, RankProfile):
+        raise ProfileError("Profile must be a RankProfile.")
```
peterrrock2:
```suggestion
        raise TypeError("Profile must be a RankProfile.")
```

src/votekit/cleaning/rank_profiles_cleaning.py
```diff
@@ -521,7 +515,93 @@ def remove_and_condense_rank_profile(
...
+
+    if isinstance(removed, Candidate) and not isinstance(removed, bool):
+        removed = [removed]
+    elif isinstance(removed, list):
+        if any(not isinstance(cand, (str, int)) or isinstance(cand, bool) for cand in removed):
+            raise TypeError("Candidates must be strings or integers within removed.")
+    else:
+        raise TypeError("removed must be a str/int candidate or a list of candidates.")
```
peterrrock2: I think that we have a function that checks if something is a valid candidate that we should use here

src/votekit/cleaning/rank_profiles_cleaning.py
```diff
@@ -521,7 +515,93 @@ def remove_and_condense_rank_profile(
...
+            [
+                profile.candidate_id_map[cand_set]
+                for cand_set in profile.candidate_id_map
+                if cand in cand_set
+            ]
+        )
+    ranking_cols = [f"Ranking_{i}" for i in range(1, profile.max_ranking_length + 1)]
+    ballots_to_remove = profile._df[ranking_cols].isin(cand_ids).any(axis=1)
+    cleaned_df = profile.df[~ballots_to_remove]
+    removed_ballot_idxs = set(profile.df[ballots_to_remove].index)
```
peterrrock2:
```suggestion
```

peterrrock2 (changes requested):
Looking good so far!

src/votekit/cleaning/rank_profiles_cleaning.py
graceg571: I added an example to the `CleanedRankProfile` class and updated the docstring to reflect the current paradigm. Will add an issue.

tests/cleaning/rank_profiles/test_remove_ballots_with_cand_rank_profile.py
graceg571: Added idempotent test that confirmed the result is the same after another application of the function. Added chaining test pin that remove ballots with a candidate and remove candidates from a ballot do not modify each other's results. All the attributes are tested for CleanProfile. Can split into a separate test for each attribute if desired.

src/votekit/cleaning/rank_profiles_cleaning.py
graceg571: Updated the docstring to reflect dropped ballots as gaps in the index. I pointed to that dropped empty ranking and zero weight ballots are recorded in the same manner.

src/votekit/cleaning/rank_profiles_cleaning.py
graceg571: Replaced all instances of `ProfileError` with `TypeError` within `rank_profiles_cleaning.py`. 1 test was broken and fixed with `TypeError`.

src/votekit/cleaning/rank_profiles_cleaning.py
graceg571: Replaced with `_validate_candidate_names`

src/votekit/cleaning/rank_profiles_cleaning.py
graceg571: The removed ballots are not considered `no_rank_altr_idxs` but are just dropped, no alteration of their ranking nor weight. Their removal is accounted for in the difference between the parent profile's index and  the cleaned profile's index. This applies to removed empty ranking and zero weight ballots, too.

src/votekit/cleaning/rank_profiles_cleaning.py
graceg571: Issue here: #392

src/votekit/pref_profile/cleaned_pref_profile.py
```diff
@@ -33,12 +49,15 @@ class CleanedRankProfile(RankProfile):
             to ``parent_profile.df``.
         no_rank_altr_idxs (set[int], optional): Set of indices of
             ballots that have no ranking as a result of cleaning. Indices are with
-            respect to ``parent_profile.df``.
+            respect to ``parent_profile.df``. A ballot has no ranking after cleaning if its ranking
+            contains only empty or tilde sets, and it had at least one valid candidated ranked
+            before cleaning.
```
peterrrock2, line 54:
```suggestion
            respect to ``parent_profile.df``. A ballot index can only be a member for the `no_rank_altr_idxs` 
            set if, after cleaning, the ballot consists of only elements of the form `frozenset()` (the 
            empty frozen set) or `frozenset({'~'})` (the set of the special character "tilde"), and it is not 
            identical to itself before applying the cleaning operation. 
```

src/votekit/utils/common_utils.py
```diff
@@ -1190,7 +1190,9 @@ def check_for_equivalent_str_int_labels(candidates: Iterable[Candidate]):
             )
 
 
-def _validate_candidate_names(candidates: Iterable[Candidate], source: object, attribute: str):
+def _validate_candidate_names(
+    candidates: Iterable[Candidate], source: Optional[object] = None, attribute: str = "candidates"
+) -> None:
```
peterrrock2, line 1195: We should use some more descriptive types her to avoid this: 

```python
   source_description = f"{source.__class__.__name__}.{attribute}" if source else f"{attribute}"
```

c.f. our Slack messages

src/votekit/cleaning/rank_profiles_cleaning.py
```diff
@@ -40,10 +40,14 @@ def _iterate_and_clean_ranking_tuples(
     cleaned_df[ranking_cols] = pd.DataFrame(cleaned_rows, index=cleaned_df.index)
 
     tilde = frozenset({"~"})
+    empty = frozenset()
     idxs = cleaned_df.index
 
     unaltr_idxs = {idx for idx, (o, c) in zip(idxs, zip(orig_rows, cleaned_rows)) if o == c}
-    no_rank_altr_idxs = {idx for idx, c in zip(idxs, cleaned_rows) if all(x == tilde for x in c)}
+    no_rank_altr_idxs = {
+        idx for idx, c in zip(idxs, cleaned_rows) if all(x == tilde or x == empty for x in c)
```
peterrrock2, line 48: Nit:

```suggestion
        idx for idx, cln_row in zip(idxs, cleaned_rows) if all(cand_set == tilde or cand_set == empty for cand_set in cln_row)
```

src/votekit/cleaning/rank_profiles_cleaning.py
```diff
@@ -61,7 +65,7 @@ def clean_rank_profile(
     clean_ranking_func: Callable[[tuple], tuple],
     remove_empty_ballots: bool = True,
     remove_zero_weight_ballots: bool = True,
-    retain_original_candidate_list: bool = True,
+    retain_original_candidate_list: bool = False,
```
peterrrock2, line 68: We should leave this as `True`

src/votekit/pref_profile/cleaned_pref_profile.py
peterrrock2, line 1: How to partition the ballots after cleaning:
- If a ballot is identical before and after cleaning, it goes in `unaltr_idxs`.
- If a ballot, as a result of cleaning, goes from having positive weight to no weight, then it goes in `no_wt_altr_idxs`.
- If a ballot, as a result of cleaning, contains only `frozenset()` and `frozenset({'~'})`, and is not a member of `unaltr_idxs`, then it goes in `no_rank_altr_idxs`
- If a ballot, as a result of cleaning, does not belong to `unaltr_idxs` and contains at least one ranking position with `frozenset(<SET_OF_CANDS>)` where `<SET_OF_CANDS>` is a non-empty set of valid candidate identifiers, then that ballot goes in `nonempy_altr_idxs`.

peterrrock2: Make sure to add unit tests for exactly these cases if they are attainable for each cleaning function. In particular, for each defined cleaning function, try to suss out all of the edge cases for the last 2 bullets.

peterrrock2 (changes requested):

src/votekit/pref_profile/cleaned_pref_profile.py
graceg571: Updated the docstring for `CleanedRankProfile` within `cleaned_pref_profile.py` to capture this partition.

This means that `_is_equiv_to_condensed` and `_is_equiv_to_remove_and_condensed` are unnecessary functions to add back `additional_unaltr_idxs` that are not identical rankings before and after cleaning but their information is not meaningfully changed. For example, a ballot that is ["A", "B", {}] where rank position 3 is empty would be ["A", "B"] after condensing. Candidates have not changed their ranking but rank position 3's empty set has been replaced with ``frozenset({'~'})` (special character tilde). Prior, `_is_equiv_to_condensed` would add this ballot's index to `unaltr_idxs` but with the updated strict paradigm, this ballot has been altered and should be a member of the `nonempty_altr_idxs` set.

Updated cleaning functions tests to test edge cases for `nonempty_altr_idxs` and `no_rank_altr_idxs`.

src/votekit/cleaning/rank_profiles_cleaning.py
graceg571: Agreed. Reverted. Will add an issue around visiting the default values for these cleaning flags.

src/votekit/pref_profile/cleaned_pref_profile.py
graceg571: Updated to follow this rule: "If a ballot, as a result of cleaning, does not belong to unaltr_idxs and contains at least one ranking position with frozenset(<SET_OF_CANDS>) where <SET_OF_CANDS> is a non-empty set of valid candidate identifiers, then that ballot goes in nonempy_altr_idxs."

src/votekit/utils/common_utils.py
graceg571: Added a dataclass for a object with attribute and `TypeAlias` for variable names to describe what we expect for the context of error messages when validating candidate's name.

src/votekit/cleaning/rank_profiles_cleaning.py
graceg571: Issue here: #396

peterrrock2: Great! Thanks!

src/votekit/utils/common_utils.py
```diff
@@ -1190,16 +1191,41 @@ def check_for_equivalent_str_int_labels(candidates: Iterable[Candidate]):
             )
 
 
-def _validate_candidate_names(candidates: Iterable[Candidate], source: object, attribute: str):
+@dataclass(frozen=True, slots=True)
+class SourceWithAttribute:
+    source: object
+    attribute: str
+
+    def __post_init(self):
```
peterrrock2: Bug: the post-init will never run
```suggestion
    def __post_init(self)__:
```

src/votekit/cleaning/rank_profiles_cleaning.py
peterrrock2, line 1: We need to update the doc strings for `condense_rank_profile` and `remove_and_condense_rank_profile`: they both still say that condensing trailing empty positions leaves a ballot "unaltered" and we changed things so that replacing trailing empty sets with padding now puts the index in `nonempty_altr_idxs`

src/votekit/utils/common_utils.py
```diff
@@ -1190,16 +1191,41 @@ def check_for_equivalent_str_int_labels(candidates: Iterable[Candidate]):
...
+        if not isinstance(self.attribute, str):
+            raise TypeError("Attribute must be a string.")
+        if not hasattr(self.source, self.attribute):
```
peterrrock2, line 1203: I am realizing that we should use `getattr_static` here since we might have some uninitialized values. The `getattr_static` function checks for declared slots, which is all we need here.

tests/cleaning/rank_profiles/test_remove_ballots_with_cand_rank_profile.py
peterrrock2: Great! We should also make sure to check steps in the chain robustly here. So doing more of 

```python
def test_remove_ballots_with_different_candidates_chaining(profile_no_ties):
    first = remove_ballots_with_cand_rank_profile("A", profile_no_ties)

    assert first.parent_profile is profile_no_ties
    assert list(first.df.index) == first.df_index_column == [1, 2]
    assert first.unaltr_idxs == {0, 1, 2, 3}

    second = remove_ballots_with_cand_rank_profile("B", first)

    assert second.parent_profile is first
    assert list(second.df.index) == second.df_index_column == [2]
    assert second.unaltr_idxs == {1, 2}

    for cleaned in (first, second):
        assert cleaned.no_rank_altr_idxs == set()
        assert cleaned.no_wt_altr_idxs == set()
        assert cleaned.nonempty_altr_idxs == set()
```

which checks not only the profille after two rounds of cleaning, but also checks the parent and that the unalter_indxs of the child refer to the immediate parent (this is very much in the vein of dotting t's and crossing i's).

The idempotence tests are good (removing "A" twice), but we should aim to cover all cases.

peterrrock2 (changes requested):
Nearly done! A couple of small things, and we should be good to go

src/votekit/utils/common_utils.py
graceg571: Good catch! Thanks! Confirmed it runs now.

src/votekit/cleaning/rank_profiles_cleaning.py
graceg571: Updated doc strings for `condense_rank_profile` and `remove_and_condense_rank_profile` to differentiate trailing empty sets that go in `nonempty_altr_idxs` and only empty sets or a mix with tilde sets that goes in `no_rank_altr_idxs`. I explained trailing empty sets are replaced with `frozenset({'~'})` while empty sets between ranking positions with candidate sets are removed and the rest of the ranking is shifted up.

src/votekit/utils/common_utils.py
graceg571: `getattr_static` would cover the case where a `source` is a `__slots__` class object and its attribute has not been initialized but is a declared slot and attribute of the object. I updated `__post_init__` to use `inspect.getattr_static` after checking if `hasattr` is True. `SlateCandMap` stores its parent (`BlocSlateConfig`) as a proxy and its attributes are accessible through `hasattr` not `getattr_static`.

PR #391, opened 2026-08-31
Add reduce_max_ranking_length option to cleaning functions

### LINKED ISSUE #322: Make `remove_and_condense_rank_profile` reduce the `max_ranking_length` of a profile.
Currently when we use `remove_and_condense_rank_profile` to remove candidates from a profile such that the total number of candidates left is less than the initial `max_ranking_length` of the profile, the resulting `max_ranking_length` will be unchanged, and thus higher than the number of candidates. This can lead to down-stream errors with functions that assume `max_ranking_length` is always at most equal to `len(profile.candidates)`, e.g. the numpy matrix converters that run fast elections, and should be fixed (although those matrix converters should also be more resilient to begin with too).

graceg571:
Closes #322 

## Summary

- Adds a `reduce_max_ranking_length: bool = False` parameter to `clean_rank_profile`, `remove_repeat_cands_rank_profile`, `remove_cand_rank_profile`, `condense_rank_profile`, and `remove_and_condense_rank_profile`
- When `True`, the returned profile's `max_ranking_length` is tightened to the max value between the maximum candidates ranked across any ballot and the minimum number of df columns needed to represent each ranking, rather than inheriting the original (potentially larger) value from the input profile. `_max_candidates_ranked` is the floor and handles candidate ties. The df columns handles repeated candidates within the profile or empty frozensets that are considered valid rankings.
- Introduces a private `_max_candidates_ranked` helper that correctly handles tied rankings by unioning candidate sets per position before counting and `_reduced_max_ranking_length` that determines the last df column with any valid candidates ranked.

## Test plan

-  Verified that `reduce_max_ranking_length=False` (default) produces identical output to previous behavior
- Verified that `reduce_max_ranking_length=True` reduces `max_ranking_length` to the minimum valid value after cleaning
- Tested with tied rankings to confirm the minimum max_ranking_length is `_max_candidates_ranked`
- Tested with repeated candidates in rankings to confirm the minimum max_ranking_length is the minimum number of columns necessary to represent all rankings

### DIFF
--- a/src/votekit/cleaning/rank_profiles_cleaning.py
+++ b/src/votekit/cleaning/rank_profiles_cleaning.py
@@ -79,7 +79,7 @@ def clean_rank_profile(
         retain_original_candidate_list (bool, optional): Whether or not to use the candidate list
             from the original profile in the new profile. If False, uses only candidates who receive
             votes. Defaults to True.
-        reduce_max_ranking_length (bool, optional): Whether or not to reduce to the profile's
+        reduce_max_ranking_length (bool, optional): Whether or not to reduce the profile's
             max_ranking_length after cleaning. If True, the max_ranking_length is set to the minimum
             number of columns necessary to represent all rankings, or to the maximum number of
             unique candidates ranked in any ballot within the profile. Whichever is higher.
@@ -168,13 +168,14 @@ def remove_repeat_cands_rank_profile(
     profile: RankProfile,
     remove_empty_ballots: bool = True,
     remove_zero_weight_ballots: bool = True,
-    retain_original_candidate_list: bool = True,
-    reduce_max_ranking_length: bool = False,
 ) -> CleanedRankProfile:
     """
     Given a profile, if a candidate appears multiple times on a ballot, keep the first instance and
-    remove any further instances. Does not condense any empty rankings as as result.
-    Only works on ranking ballots, not score ballots.
+    remove any further instances. The repeated candidate is replaced with an empty set if ranked
+    alone or removed from the set if tied. Does not condense any empty rankings as as result.
+    Only works on ranking ballots, not score ballots. The profile's candidates list and
+    max_ranking_length are preserved. Can condense the ballots after removing repeated candidate(s)
+    by calling ``condense_rank_profile`` on the result.
 
     Wrapper for clean_rank_profile.
 
@@ -184,13 +185,6 @@ def remove_repeat_cands_rank_profile(
             ranking or scores as a result of cleaning. Defaults to True.
         remove_zero_weight_ballots (bool, optional): Whether or not to remove ballots that have no
             weight as a result of cleaning. Defaults to True.
-        retain_original_candidate_list (bool, optional): Whether or not to use the candidate list
-            from the original profile in the new profile. If False, uses only candidates who receive
-            votes. Defaults to True.
-        reduce_max_ranking_length (bool, optional): Whether or not to reduce to the profile's
-            max_ranking_length after cleaning. If True, the max_ranking_length is set to the max
-            candidates ranked of any ballot within the profile which is the minimum allowed.
-            Defaults to False.
 
     Returns:
         CleanedRankProfile: A cleaned ``RankProfile``.
@@ -204,8 +198,8 @@ def remove_repeat_cands_rank_profile(
         remove_repeat_cands_from_ranking_row,
         remove_empty_ballots,
         remove_zero_weight_ballots,
-        retain_original_candidate_list,
-        reduce_max_ranking_length,
+        retain_original_candidate_list=True,
+        reduce_max_ranking_length=False,
     )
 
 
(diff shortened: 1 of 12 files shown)


fetachino:
Potential behavior question: the new reduce_max_ranking_length paths call _max_candidates_ranked(profile), using the original profile after cleaning has already removed candidates, empty ballots, or zero-weight ballots. In those cases the returned max_ranking_length may remain larger than the cleaned data requires, so the option may not actually reduce the result. Should this be computed from the cleaned profile instead (with a regression test covering a ballot removed during cleaning), unless preserving the original profile width is intentional?

graceg571:
I did not mean to call `_max_candidates_ranked` on the original profile but the `cleaned_profile`. Good catch!

peterrrock2 (changes requested):
Okay functions look implemented correctly here, but I think there are a couple of small integration checks missing.

## Breaking API changes

The functions `condense_rank_profile` and `remove_repeat_cands_rank_profile` have changes to their public API that we should be careful about. Renaming variables is not that big of an issue, but changes to calls for positional arguments should be treated carefully. We should preserve the position and behavior of `retain_original_candidate_list`, deprecate it if we intend to remove it, and make `reduce_max_ranking_length` keyword-only.

For example, this existing call:

```python
condense_rank_profile(profile, True, True, True)
```

previously preserved the candidate list and ranking width. Now the final `True` enables width reduction. Calls using `retain_original_candidate_list` as a keyword also raise `TypeError`, so we should preserve those during any deprecation period.

## What happens when you remove everything

There is an unintuitive interaction when you call `remove_and_condense_rank_profile` on a profile which, as a result of cleaning, will have all of its ballots removed.

```python
profile = RankProfile(
    ballots=[RankBallot(ranking=[{"A"}])],
    max_ranking_length=4,
)
cleaned = remove_and_condense_rank_profile(
    "A", profile, reduce_max_ranking_length=True
)
```

`_reduced_max_ranking_length(cleaned_df)` returns zero, but the constructor infers the width from the retained dataframe columns. The result, therefore, still has `max_ranking_length == 4`, despite containing no ballots. The constructor needs to preserve an explicitly requested width of zero instead of treating it as a request to infer the width.

There is also a related case when we retain the empty ballot:

```python
remove_and_condense_rank_profile(
    "A",
    profile,
    remove_empty_ballots=False,
    reduce_max_ranking_length=True,
)
```

This raises `ProfileError` because the dataframe constructor rejects a zero ranking width when rows remain. The constructor also needs to support an explicit zero width when empty ballot rows are retained.

I think that the fix here is to use the same constructor rules for ballot and dataframe inputs: `max_ranking_length=None` means infer the width, `0` means explicitly zero ranking positions, and a positive integer means preserve that limit even when there are no ballots. 

## Adding profiles with extra ranking columns

This one is a bit subtle and I only stumbled on it messing about. We reduce `max_ranking_length`, but the dataframe still contains the original number of ranking columns. That is fine as long as the extra columns contain only padding: `max_ranking_length` has meaning independently of the dataframe's physical width, including determining the default Borda score vector.

```python
profile = RankProfile(
    ballots=[RankBallot(ranking=[{"A"}, {"B"}])],
    max_ranking_length=4,
)
cleaned = condense_rank_profile(profile, reduce_max_ranking_length=True)
```

Here, `cleaned.max_ranking_length` is `2`, but its dataframe still contains `Ranking_1` through `Ranking_4`. Currently, adding the profiles raises:

```python
cleaned + profile
# ValueError: cannot insert Ranking_3, already exists
```

Addition tries to pad the shorter profile to the wider profile's declared width, but those columns already exist.

We should update `_sum_rank_profiles` to insert only the required ranking columns that are actually missing. There is no need to require cleaning to drop the extra padding columns. A test that reduces a profile and then adds it to a wider profile should verify that addition succeeds and preserves ballots and weights.

graceg571:
## Breaking API Changes
This makes sense. For functions `condense_rank_profile` and `remove_repeat_cands_rank_profile`, I was thinking that `retain_original_candidate_list` would have no effect similar to `reduce_max_ranking_length` being applicable to only condensing functions. However, the fallback for those functions is to set the candidates list to only candidates with a vote casted. I reverted the API Changes and made `reduce_max_ranking_length` a keyword only argument for `clean_rank_profile` and all other profile cleaning functions. `reduce_max_ranking_length` makes sense for condense cleaning functions but can have an effect on remove candidate cleaning functions if the original parent has a `max_ranking_length` greater than it's minimum allowed (max of `max_candidates_ranked` or minimum number of ranking columns to represent any ranking). 

## Cleaned profile has no leftover ballots or only null ballots
If you remove all ballots, what is the max_ranking_length? Should be zero but the constructor sets it to the df's number of ranking columns per `_find_max_ranking_length` because it considers a `max_ranking_length` of zero or None as unset. `_find_max_ranking_length` is updated to infer the width if `max_ranking_length` is None. However, `max_ranking_length` is immediately set to 0 if it's `None` so 0 and None are conflated for both profiles initialized with ballots or df. I wrote an issue to pull apart profiles with unset, or `None`, `max_ranking_length` from ones with intentional `max_ranking_length=0` per *update with issue once written*.

A `ProfileError` should not be thrown for a profile with a `max_ranking_length=0` if the df's rows contains rankings composed of entirely ``frozenset({'~'})``. `__validate_init_rank_df_params` now only throws the `ProfileError` for a `max_ranking_length` of 0 if the df contains non-null ballot rows.

When a cleaned profile has no leftover ballots, the df is empty with no rows but still has its columns defined (at least `VoterSet` and `Weight`) but its ballots list is empty. This is equivalent to passing an empty list to `ballots`.
When a cleaned profile has only null ballots, the df is not empty; it's composed of rows with rankings of all `frozenset('~')` with their weight and voter set. The cleaned profile `ballots` attribute returns a list of ballots with no ranking but a `weight` and `voter_set` if set.  This is equivalent to passing a list of ballots with `ranking=(),` to `ballots`. Confirmed the `max_ranking_length`, `ballots`, and `num_ballots` were the same amongst the equivalent df and ballots cases.
In both cases, the `max_ranking_length` should be zero but the cleaned profile's df has ranking columns equal to its parent profile's `max_ranking_length`. This is related to issue #397 where the `max_ranking_length` can be less than the number of ranking columns of the passed df without throwing an error. Yet, a shorter `max_ranking_length` than a df's width is not an issue if the extra ranking columns are just padding (i.e. all `frozenset({'~'})`).

## Adding profiles with extra ranking columns
Noticed this too per issue #397 but agree this is fine if the extra columns are only padding. Updated sum to handle this and added a test to `test_sum_profiles` that sums a profile with reduced profile where its df width is greater than its `max_ranking_length`.

I also noticed that `PreferenceProfile` expects a non-optional int `max_ranking_length` while `RankProfile` has `max_ranking_length` defined as optional int. Only an issue for the static type checker.

